PR #3, opened 2025-01-23
feat: Byte decomposition example

storojs72:
This PR adds example of using LinearCombination column in order to implement byte decomposing.


storojs72:
Sorry, missed the upstream branch

PR #6, opened 2025-02-06
[field] Add TowerField::min_tower_level(self), and use it to derive ArithExpr tower_level from its constants

bergkvist:
In contrast to `TowerField::TOWER_LEVEL`, `TowerField::binary_tower_level(self)` returns the smallest tower level that can fit the current value. This can be useful for shrinking field values to the smaller container that fits them, for the purpose of making arithmetic operations (in particular multiplication) cheaper

Note that depending on the tower family, not all tower level values are valid.
- For BinaryField, tower levels 0, 1, 2, 3, 4, 5, 6, 7 are valid
- For AESTowerField, tower levels 0, 3, 4, 5, 6, 7 are valid
- For Polyval, tower levels 0, 7 are valid

### DIFF
--- a/crates/field/src/binary_field.rs
+++ b/crates/field/src/binary_field.rs
@@ -45,6 +45,13 @@ where
 	/// Currently for every tower field, the canonical field is Fan-Paar's binary field of the same degree.
 	type Canonical: TowerField + SerializeBytes + DeserializeBytes;
 
+	/// Returns the smallest valid `TOWER_LEVEL` in the tower that can fit the same value.
+	///
+	/// Since which `TOWER_LEVEL` values are valid depends on the tower,
+	/// `F::Canonical::from(elem).binary_tower_level()` can return a different result
+	/// from `elem.binary_tower_level()`.
+	fn min_tower_level(self) -> usize;
+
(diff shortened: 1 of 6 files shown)


crates/field/src/binary_field.rs
```diff
@@ -45,6 +45,13 @@ where
 	/// Currently for every tower field, the canonical field is Fan-Paar's binary field of the same degree.
 	type Canonical: TowerField + SerializeBytes + DeserializeBytes;
 
+	/// Returns the smallest valid `TOWER_LEVEL` in the tower that can fit the same value.
+	///
+	/// Since which `TOWER_LEVEL` values are valid depends on the tower,
+	/// `F::Canonical::from(elem).binary_tower_level()` can return a different result
+	/// from `elem.binary_tower_level()`.
+	fn binary_tower_level(self) -> usize;
```
jimpo-ulvt, line 53: Please rename to `min_tower_level`

jimpo-ulvt (changes requested):
Looks good, with one renaming request

onesk (approved):

jimpo-ulvt (approved):

PR #7, opened 2025-02-06
[serialization] Add canonical serialize/deserialize traits + derive macros

bergkvist:
Introduces the following traits:
- SerializeCanonical (which replaces most uses of SerializeBytes)
- DeserializeCanonical (which replaces most uses of DeserializeBytes)

Conveniently, this also comes with proc-macros for deriving these traits for an arbitrary struct/enum (unions are not supported).


Anex007 (approved):
Looks good to me, Now I wonder if there's a reason for keeping `DeserializeBytes/SerializeBytes` in the utils crate, as we seem to use this only internally inside the field crate.

Anex007 (dismissed):
Nice

PR #12, opened 2025-02-07
Implement PackedField::unzip

GraDKh:


crates/field/src/arch/portable/packed_arithmetic.rs
```diff
@@ -37,6 +37,18 @@ where
 
 		(c, d)
 	}
+
+	/// Transpose with the given bit size
```
jimpo-ulvt, line 41: Maybe add to the docstring that this is used to implement `PackedField::unzip`

jimpo-ulvt (approved):

PR #13, opened 2025-02-10
[refac]: rm some useless checked_log_2 calls

tcoratger:


crates/ntt/src/additive_ntt.rs
```diff
@@ -51,11 +51,7 @@ pub trait AdditiveNTT<P: PackedField> {
 		PE: RepackedExtension<P>,
 		PE::Scalar: ExtensionField<P::Scalar>,
 	{
-		if !PE::Scalar::DEGREE.is_power_of_two() {
-			return Err(Error::PowerOfTwoExtensionDegreeRequired);
-		}
-
-		let log_batch_size = log2_strict_usize(PE::Scalar::DEGREE);
```
tcoratger, line 58: Here I'm not sure why we needed these check because PE::Scalar::DEGREE is already hard defined as a power of 2 by construction no?

jimpo-ulvt: This is a change of behavior though, switching from `log(DEGREE)` to `log(LOG_DEGREE)`.

tcoratger: Yes of course sorry, just fixed

jimpo-ulvt (approved):

PR #14, opened 2025-02-10
[core]: simplify merkle tree `verify_opening`

tcoratger:


crates/core/src/merkle_tree/scheme.rs
```diff
@@ -108,42 +108,36 @@ where
...
 		}
 
-		let leaf_digest = hash_field_elems::<_, H>(values);
-		let branch = proof.read_vec(tree_depth - layer_depth)?;
-
-		let mut index = index;
-		let root = branch.into_iter().fold(leaf_digest, |node, branch_node| {
-			let next_node = if index & 1 == 0 {
-				self.compression.compress([node, branch_node])
+		let mut root = hash_field_elems::<_, H>(values);
```
djadjka, line 128: If I had to choose between root and leaf_digest, I would say leaf_digest is the better option.

tcoratger: Fixed

djadjka:
LGTM

jimpo-ulvt (approved):

PR #15, opened 2025-02-10
[core]: Statically compile exponentiation in ArithCircuitPoly

tcoratger:


crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -50,10 +50,26 @@ fn circuit_steps_for_expr<F: Field>(
...
+				let exp = *exp;
+				let base_arg = to_circuit_inner(base, result);
+
+				if exp == 1 {
+					// x^1 = x
+					return base_arg;
+				}
+
+				// Start with x^2 = x * x
+				result.push(CircuitStep::Mul(base_arg, base_arg));
```
jimpo-ulvt, line 63: Please introduce an actual `Square` step, as squaring in binary fields is much cheaper than regular multiplication.

jimpo-ulvt (changes requested):

GraDKh:
TBH the most optimal solution would be to introduce specialized steps for powers 2-4. Otherwise there will be negative effect of having more steps (which means more extra memory reads/writes).
BTW are there any benchmark results to see what is the effect?

jimpo-ulvt:
> TBH the most optimal solution would be to introduce specialized steps for powers 2-4. Otherwise there will be negative effect of having more steps (which means more extra memory reads/writes). BTW are there any benchmark results to see what is the effect?

I agree that adding an optimization pass that combines two square steps into a single Pow4 step would probably be a good optimization, though I think it can be done as a follow-up.

crates/core/src/polynomial/arith_circuit.rs
tcoratger: Yes it seems logical to me, as the `CircuitStep::Square` did not exist I did not know if it was normal or if it had to be added.

- I just modified, tell me what you think, I added tests to be sure of my construction but as it is the first time I touch this part, I may have made mistakes, let me know.

- If I am not mistaken `CircuitStep::Pow` is no longer used at all and can therefore be deleted right?

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -50,10 +50,34 @@ fn circuit_steps_for_expr<F: Field>(
...
+				// Apply square and minimize multiplications
+				while current_exp * 2 <= exp {
+					result.push(CircuitStep::Square(acc));
+					acc = CircuitStepArgument::Expr(CircuitNode::Slot(result.len() - 1));
+					current_exp *= 2;
+				}
+
+				// If remaining exponentiation, multiply by base_arg
+				if current_exp < exp {
+					for _ in current_exp..exp {
```
jimpo-ulvt, line 74: This is doing too many multiplications. Use the [square and multiply](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) algorithm.

jimpo-ulvt (changes requested):

crates/core/src/polynomial/arith_circuit.rs
tcoratger: Should be fixed let me know if you feel this is the best approach. I added a bunch of tests to validate this tricky impl locally so I just kept them here to be safe but can delete some of them if needed.

jimpo-ulvt: I think the code will be simpler if you iterate the bits high to low, like [here](https://github.com/IrreducibleOSS/binius/blob/main/crates/field/src/field.rs#L89). Then you only need to track one accumulator with no special cases, instead of a separate accumulator of squared bases.

tcoratger: Mhhh due to the `base_expr` stuff we need to modify a bit the algorithm you mention I think, do you mean this?

```rust
ArithExpr::Pow(base, exp) => {
  let mut acc = to_circuit_inner(base, result);
  let base_expr = acc;
  
  for i in (0..64).rev() {
	  if i < exp.ilog2() as u32 {
		  result.push(CircuitStep::Square(acc));
		  acc = CircuitStepArgument::Expr(CircuitNode::Slot(result.len() - 1));
	  }
	  if (exp >> i) & 1 != 0 && i != exp.ilog2() as u32 {
		  result.push(CircuitStep::Mul(acc, base_expr));
		  acc = CircuitStepArgument::Expr(CircuitNode::Slot(result.len() - 1));
	  }
  }
  
  acc
}
```

I think this works and only has `acc` as accumulator.

jimpo-ulvt: You can just iterate `(0..highest_bit).rev()` and avoid the conditionals, I think.

tcoratger: fixed, sorry for the back and forth, I think that now I better understood how the circuit construction works

jimpo-ulvt (approved):

PR #20, opened 2025-02-12
Karatsuba interpolation in sumchecks

onesk:
This MR adds support for the Karatsuba infinity point when interpolating, where we take a binary subspace evaluation domain and replace the third point (after zero and one) with an infinity: 

$$
f(z, xs) = f(0, xs) + z (f(1,xs) - f(0,xs)) 
$$
$$
f(\infty, xs) = lim_{z \rightarrow \infty} \frac{f(z,xs)}{z} = f(1,xs) - f(0, xs)
$$

This saves a multiplication in one point (which is a large-by-large in POLYVAL case) and allows sumchecking quadratic constraints without using multiplications for interpolation at all. There is a caveat though - we cannot use the original composition polynomial $C$ in this point, and should use $lim_{z \rightarrow \infty} \frac{C(f(z, xs))}{{z^{deg C}}} z^{deg C}$ instead. Luckily having `expression()` in `CompositionPolyOS` and `ArithExpr`/`ArithCircuitPoly` makes this pretty straightforward.

This optimization is prover-only and provides consistent 10-15% improvement on single threaded GPA sumchecks, and should never cause a regression, though its benefit is best felt on domains not capable of small by large multiplication (like POLYVAL) and composition polynomials which get simpler when only the highest degree term is concerned.

It is however somewhat invasive in a sense that sumcheck evaluator should be acutely aware of the infinity point existence - while we are likely to only support regular & equality indicator sumchecks going forward, this is still code to maintain. I would like a second opinion whether it's worth merging this.

### DIFF
--- a/crates/core/src/protocols/sumcheck/prove/zerocheck.rs
+++ b/crates/core/src/protocols/sumcheck/prove/zerocheck.rs
@@ -564,13 +564,16 @@ where
 			round == 0 && matches!(self.first_round, RegularFirstRound::SkipCube);
 		let coeffs = if skip_cube_first_round {
 			let evaluators = izip!(&self.compositions, &self.domains)
-				.map(|(composition, interpolation_domain)| ZerocheckFirstRoundEvaluator {
-					composition,
-					composition_highest_degree_only: ArithCircuitPoly::new(
-						composition.expression().highest_degree_only().1,
-					),
-					interpolation_domain,
-					partial_eq_ind_evals: &self.partial_eq_ind_evals,
+				.map(|(composition, interpolation_domain)| {
+					let composition_at_infinity =
+						ArithCircuitPoly::new(composition.expression().highest_degree_only().1);
+
+					ZerocheckFirstRoundEvaluator {
+						composition,
+						composition_at_infinity,
+						interpolation_domain,
+						partial_eq_ind_evals: &self.partial_eq_ind_evals,
+					}
 				})
 				.collect::<Vec<_>>();
 			let evals = self.state.calculate_round_evals(&evaluators)?;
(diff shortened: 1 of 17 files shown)


crates/hal/src/sumcheck_evaluator.rs
```diff
@@ -23,6 +23,7 @@ pub trait SumcheckEvaluator<P: PackedField, Composition> {
 		&self,
 		subcube_vars: usize,
 		subcube_index: usize,
+		is_infinity_point: bool,
```
jimpo-ulvt, line 26: May need to expose `composition_at_infinity` as a method as well.

jimpo-ulvt (commented):
I think this is a good feature, even if it's only a 10% speedup right now. Hopefully as other things get optimized it becomes more significant for GPA.

crates/hal/src/sumcheck_evaluator.rs
jimpo-ulvt: Also, please add documentation explaining what evaluation at infinity means and referencing a source for it.

crates/core/src/protocols/sumcheck/prove/zerocheck.rs
```diff
@@ -620,14 +624,15 @@ where
 	FDomain: Field,
 {
 	composition: &'a Composition,
+	composition_highest_degree_only: ArithCircuitPoly<P::Scalar>,
```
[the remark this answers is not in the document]
onesk: Agree, renamed.

crates/core/src/protocols/sumcheck/prove/zerocheck.rs
```diff
@@ -788,6 +815,13 @@ where
 
 		round_evals.insert(0, zero_evaluation);
 
+		if round_evals.len() > 3 {
+			// Reorder eval at infinity point from index 2 to the last position
```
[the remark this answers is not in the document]
onesk: `SumcheckRoundCalculator` orders "trivial" ($0$, $1$, $\infty$) points at the beginning of the domain because the interpolation loop is the outer one, thus such ordering enables savings even under mixed composition degrees. Left a comment on that.

crates/hal/src/sumcheck_evaluator.rs
onesk: Added a link to Wikipedia on `EvaluationDomain` doc comment and referenced it in `SumcheckEvaluator`.

As for exposing `composition_at_infinity`, this is the first approach that comes to mind but sadly it doesn't play well with dyn compatibility of the `CompositionPolyOS`. I think using `expression` is fine for that purpose.

jimpo-ulvt: Mmm, I see.

jimpo-ulvt (approved):
Please rebase and address the one comment.

crates/math/src/arith_expr.rs
```diff
@@ -60,6 +60,33 @@ impl<F: Field> ArithExpr<F> {
 		}
 	}
 
+	/// Return a new arithmetic expression that contains only the terms of highest degree
+	/// (useful for interpolation at Karatsuba infinity point).
+	pub fn highest_degree_only(&self) -> (usize, Self) {
```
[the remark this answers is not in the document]
onesk: Introduced the `leading_term()` method.

jimpo-ulvt (approved):

PR #21, opened 2025-02-12
Byte-sliced fields changes

GraDKh:
In this MR initially transposed byte sliced fields are added. They show almost the same (~20% less) multiplication throughput while keeping the per-element access fast (5-10x faster than byte-sliced fields).

However now the idea is to implement transposed byte-sliced fields in a separate PR, although other changes in this PR are useful in other places:

1. Refactor a bit `TowerLevels` to remove packed field parameter from the `TowerLevel` to the `Data` associated type. This also makes generic bounds a bit more clean, since TowerLevel itself doesn't depend on a concrete packed field type.
2. Add support of byte-sliced fields with arbitrary register size, i.e. 128b, 256b, 512b.
3. Add shifts and unpack low/high within 128-bit lanes to `UnderlierWithBitOps`. This allows implementing transposition  in an efficient way.
4. Add the transparent implementation of `UnderlierWithBitOps` for `PackedScaledUnderlier` as we need it to re-use `PackedScaledField`.


djadjka (approved):

jimpo-ulvt:
As discussed, please remove the new packed field type and leave the other helper code

PR #27, opened 2025-02-14
[circuits] Add test_circuit helper

bergkvist:


crates/circuits/src/builder/test_utils.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use binius_core::constraint_system::{channel::Boundary, validate::validate_witness};
+
+use super::{types::F, ConstraintSystemBuilder};
+
+pub fn test_circuit(
+	build_circuit: fn(&mut ConstraintSystemBuilder) -> Result<Vec<Boundary<F>>, anyhow::Error>,
+) -> Result<(), anyhow::Error> {
+	let allocator = bumpalo::Bump::new();
+	let mut builder = ConstraintSystemBuilder::new_with_witness(&allocator);
```
jimpo-ulvt, line 11: It would be better & more realistic for the prover and verifier to build separate constraint systems with the same closure. Like the prover builds with the witness, then takes the witness. Then the verifier builds again without a witness, getting the advice and witness from the prover.

jimpo-ulvt (approved):

PR #37, opened 2025-02-18
[benchmarks] Added benchmarks for commitment and flushes

Anex007:
This MR adds benchmarks for polynomial commitment and GKR GPA in terms of flushes


crates/core/benches/poly_commit.rs
```diff
@@ -0,0 +1,141 @@
...
+
+const LOG_SIZE: usize = 20;
+
+fn bench_poly_commit(c: &mut Criterion) {
+	let (fri_params, merkle_prover, committed_multilins) = create_poly_commit(LOG_SIZE);
+	let mut group = c.benchmark_group("Polynomial Commitment");
+	group.throughput(Throughput::Bytes(((1 << LOG_SIZE) * BinaryField128b::N_BITS / 8) as u64));
+	group.bench_function(BenchmarkId::new("log_size", LOG_SIZE), |b| {
+		b.iter(|| {
+			let rs_code = ReedSolomonCode::new(
```
jimpo-ulvt, line 68: This shouldn't create the RS code in the iter loop

crates/core/benches/poly_commit.rs
```diff
@@ -0,0 +1,141 @@
...
+				fri_params.rs_code().log_dim(),
+				fri_params.rs_code().log_inv_rate(),
+				&NTTOptions {
+					precompute_twiddles: true,
+					thread_settings: ThreadingSettings::SingleThreaded,
+				},
```
GraDKh, line 74: Should we have benches with different options here?
I have doubts about multi-threaded (unless for big amount of data), but seems reasonable to test with and without precomputed twiddles

GraDKh (approved):

crates/core/benches/poly_commit.rs
Anex007: In my machine `M3 pro` and `c7a.2xlarge`  I did not see a noticable difference with and without precomputed twiddles in both single threaded and multithreaded runs

jimpo-ulvt (changes requested):
The time required for `poly_commit` seems to be more than the sum of the binary merkle tree building and NTT combined, judging from throughput numbers. Could you please double check that and if you observe the same thing, profile where else `poly_commit` is spending time?

crates/ntt/benches/additive_ntt.rs
```diff
@@ -144,5 +148,42 @@ fn bench_inverse_transform(c: &mut Criterion) {
...
+				.precompute_twiddles()
+				.multithreaded();
+			group.bench_function(BenchmarkId::new("multithread/precompute", &params), |b| {
+				b.iter(|| ntt.forward_transform(&mut data, 0, log_batch_size));
+			});
+		}
+	}
+}
+
+criterion_group!(ntt, bench_forward_transform, bench_inverse_transform, bench_large_transform);
```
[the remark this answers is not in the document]
Anex007: Thats a good idea

Anex007:
I've updated the input parameters so that the newly added benchmarks for merkle tree building and NTT are consistent with the polynomial commitment throughputs.

jimpo-ulvt (approved):

PR #38, opened 2025-02-19
feat: Blake3 compression

storojs72:
This PR completes set of gadgets to implement Blake3 [compression](https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs#L74) circuit. [Permutation](https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs#L66), which is a part of compressing, is not yet entirely constrained - @onesk suggested to use channel balancing constraint here (according to the [following](https://gitlab.com/IrreducibleOSS/binius/-/merge_requests/601/diffs#e209d5579974b8e6590048359fdbc677201f6c83_0_895) example in KeccakSponge), which I'm going to address for the Blake3 case in a subsequent PR.

### DIFF
--- a/crates/circuits/src/blake3.rs
+++ b/crates/circuits/src/blake3.rs
@@ -7,12 +7,19 @@ use binius_utils::checked_arithmetics::checked_log_2;
 use crate::{
 	arithmetic,
 	arithmetic::Flags,
+	bitwise,
 	builder::{types::F, ConstraintSystemBuilder},
+	sha256::u32const_repeating,
+	unconstrained::fixed_u32,
 };
 
 type F1 = BinaryField1b;
 const LOG_U32_BITS: usize = checked_log_2(32);
+const CHAINING_VALUE_LEN: usize = 8;
 const BLAKE3_STATE_LEN: usize = 16;
+const MSG_PERMUTATION: [usize; BLAKE3_STATE_LEN] =
+	[2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8];
+const IV_0_4: [u32; 4] = [0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A];
 
 // Gadget that performs two u32 variables XOR and then rotates the result
 fn xor_rotate_right(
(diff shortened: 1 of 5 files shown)


crates/circuits/src/blake3.rs
```diff
@@ -72,7 +81,7 @@ pub fn blake3_g(
 	mx: OracleId,
 	my: OracleId,
 	log_size: usize,
-) -> Result<[OracleId; 4], anyhow::Error> {
+) -> Result<(OracleId, OracleId, OracleId, OracleId), anyhow::Error> {
```
onesk, line 84: nit: why not an array (like it was), it pattern matches like a tuple and the type signature is subjectively better this way?

crates/circuits/src/blake3.rs
```diff
@@ -95,28 +104,160 @@ pub fn blake3_g(
...
+
+	builder.pop_namespace();
+
+	Ok([
+		s0, s1, s2, s3, s4, s5, s6, s7, s8, s9, s10, s11, s12, s13, s14, s15,
+	])
+}
+
+// TODO: implement permutation constraining
+pub fn permute(block_words: &[OracleId]) -> [OracleId; BLAKE3_STATE_LEN] {
```
onesk, line 153: nit: this can actually be an array reference `&[OracleId; BLAKE3_STATE_LEN]` as well instead of slice, making the assertion compile time

crates/circuits/src/unconstrained.rs
```diff
@@ -47,11 +47,11 @@ where
...
-	let rng = builder.add_committed(name, log_size, FS::TOWER_LEVEL);
+	let fixed = builder.add_committed(name, log_size, FS::TOWER_LEVEL);
 
 	if let Some(witness) = builder.witness() {
 		witness
-			.new_column::<FS>(rng)
+			.new_column::<FS>(fixed)
 			.as_mut_slice::<u32>()
 			.into_par_iter()
 			.zip(value.into_par_iter())
```
onesk, line 57: nit: plural `values`

crates/circuits/src/blake3.rs
```diff
@@ -7,11 +7,19 @@ use binius_utils::checked_arithmetics::checked_log_2;
 use crate::{
 	arithmetic,
 	arithmetic::Flags,
+	bitwise,
 	builder::{types::F, ConstraintSystemBuilder},
+	sha256::u32const_repeating,
```
onesk, line 12: better to factor out `u32const_repeating` into the library code instead of referencing another hash gadget - there is nothing SHA256-specific in constant `u32` repeating oracles :)

crates/circuits/src/blake3.rs
```diff
@@ -95,28 +104,160 @@ pub fn blake3_g(
...
+	let mut state = [OracleId::MAX; BLAKE3_STATE_LEN];
+	state[0..8].copy_from_slice(chaining_value);
+
+	let iv_oracles =
+		IV_0_4.map(|val| u32const_repeating(log_size, builder, val, "blake3_iv").unwrap());
+
+	state[8..12].copy_from_slice(&iv_oracles);
+
+	state[12] =
+		fixed_u32::<F1>(builder, "counter_low", log_size, vec![counter_low; 1 << log_size])?;
```
onesk, line 191: all four of these oracles are `u32_const_repeating`, no need to commit them.

crates/circuits/src/blake3.rs
```diff
@@ -95,28 +104,160 @@ pub fn blake3_g(
...
+
+	Ok([
+		s0, s1, s2, s3, s4, s5, s6, s7, s8, s9, s10, s11, s12, s13, s14, s15,
+	])
+}
+
+// TODO: implement permutation constraining
+pub fn permute(block_words: &[OracleId]) -> [OracleId; BLAKE3_STATE_LEN] {
+	assert_eq!(block_words.len(), BLAKE3_STATE_LEN);
+	let mut permuted = [OracleId::MAX; BLAKE3_STATE_LEN];
```
onesk, line 155: this can be an `array::from_fn`

crates/circuits/src/blake3.rs
```diff
@@ -95,28 +104,160 @@ pub fn blake3_g(
...
 
-	Ok([a2, b2, c2, d2])
+	Ok((a2, b2, c2, d2))
+}
+
+// Gadget for Blake3 round function
+pub fn round(
+	builder: &mut ConstraintSystemBuilder,
+	name: impl ToString,
+	state: &[OracleId],
```
onesk, line 114: same argument about using an array reference vs slice

onesk (changes requested):
Some code style remarks

crates/circuits/src/blake3.rs
storojs72: Nice. Thanks!

crates/circuits/src/blake3.rs
storojs72: That's because `g` is used in the `round` function and output oracles of previous `g` is used as input for the next one, so we can use:

```
let (s0, s4, s8, s12) = g(builder, "mix-columns-0", state[0], state[4], state[8], state[12], m[0], m[1], log_size)?;
. . . 
. . .
let (s0, s5, s10, s15) = g(builder, "mix-diagonals-0", s0, s5, s10, s15, m[8], m[9], log_size)?;
let (s3, s4, s9, s14) = g(builder, "mix-diagonals-3", s3, s4, s9, s14, m[14], m[15], log_size)?;
```

If we use array as output, then we will tentatively have:
```
let s_mc_0 = g(builder, "mix-columns-0", state[0], state[4], state[8], state[12], m[0], m[1], log_size)?;
. . .
. . .
let s_md_0 = g(builder, "mix-diagonals-0", s_mc_0[0], s5, s10, s15, m[8], m[9], log_size)?;
let s_md_3 = g(builder, "mix-diagonals-3", s3, s_mc_0[1], s9, s14, m[14], m[15], log_size)?; 
 ```
 
 which looks more confusing...

**Comment by storojs72** [comment]
We discussed out-of-PR that this kind of permutations like in Blake3 - where input has the same length as output and essentially indices are just reshuffled - don't require special constraining.

File: crates/circuits/src/blake3.rs (hunk 1 above)
Comment (onesk) [reply crates/circuits/src/blake3.rs:85]: You should be perfectly fine doing that:

```rust
let [a, b, c, d] = g(...)?;
let [e, f, g, h] = g(..., [a, b, c, d])?;
```

onesk (approved):
Looks good to me, one last stylistic comment which is up to you.

crates/circuits/src/blake3.rs
storojs72: I'm going to add an example file in the next trivial PR. Let me include this there.

PR #39, opened 2025-02-19
Generator exponentiation batch prove/verify 

djadjka:
This PR contains general code for exponentiation of both static and dynamic generators. Batching starts from the last exponent bit, with each batch containing claims for the same evaluation point. When all claims share the same evaluation point, this approach allows using a single GKR sumcheck prover for each layer.


crates/core/src/composition/index.rs
```diff
@@ -134,6 +134,77 @@ where
 	})
 }
 
+#[derive(Debug)]
+pub enum ComplexIndexComposition<C> {
```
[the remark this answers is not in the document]
onesk: On a second thought, there is no need for this struct at all if you leverage the fact that `SumcheckProver` is a dyn-compatible trait.

onesk (changes requested):
I admire your effort to coalesce everything into a single sumcheck prover on each layer, but I'm afraid additional machinery makes the code pretty much inpenetrable for someone trying to understand it. I would suggest to start with these simplifications:
 1) Leverage the dyn compatibility of `SumcheckProver` and create "narrow" provers containing only individual claims - this should hopefully remove some of the batching complexity overhead
 2) Factor the four way decision tree "generator vs dynamic base" x "first/last vs inner round" into separate functions

 Overall the effect of this should be comparable to what you achieve when refactoring state machines into async/await. In the ideal case instead of a single protocol object handling all subcases you should arrive at separate code paths for static & dynamic exponentiations at least. After that we should probably join forces in the follow up in an effort to unify the GKR logic between GPA and int muls.

crates/core/src/protocols/gkr_exponentiation/oracles.rs
```diff
@@ -0,0 +1,134 @@
...
+	protocols::{evalcheck::EvalcheckMultilinearClaim, gkr_gpa::LayerClaim},
+};
+
+#[derive(Clone)]
+pub struct ExponentiationClaimMeta<F> {
+	evals: Vec<F>,
+	pub with_dynamic_base: Vec<bool>,
+}
+
+pub fn get_evals_and_with_dynamic_base_from_witnesses<P, PBase>(
```
jimpo-ulvt, line 23: This is not a good function signature. "and with" doesn't read correctly in english grammar.

crates/core/src/protocols/gkr_exponentiation/oracles.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use super::{common::ExponentiationClaim, error::Error, witness::BaseExponentWitness};
+use crate::{
+	fiat_shamir::Challenger,
+	oracle::{MultilinearOracleSet, OracleId},
+	protocols::{evalcheck::EvalcheckMultilinearClaim, gkr_gpa::LayerClaim},
+};
+
+#[derive(Clone)]
+pub struct ExponentiationClaimMeta<F> {
```
jimpo-ulvt, line 18: Why is this batched as a collection of data instead of independent structs?

crates/core/src/protocols/gkr_exponentiation/oracles.rs
jimpo-ulvt: Is this method used? Where?

crates/core/src/protocols/gkr_exponentiation/provers.rs
```diff
@@ -0,0 +1,329 @@
...
+	) -> Result<Self, Error> {
+		if witness.base.is_none() {
+			bail!(Error::IncorrectWitnessType);
+		}
+
+		Ok(Self(ExponentiationCommonProver::new(witness, claim.clone())))
+	}
+}
+
+pub struct ProverLayerClaimMeta<'a, P: PackedField> {
```
jimpo-ulvt, line 241: What is this? Why is claim an `Option`?

crates/core/src/protocols/gkr_exponentiation/witness.rs
```diff
@@ -0,0 +1,255 @@
...
+use bytemuck::zeroed_vec;
+
+use super::error::Error;
+use crate::{protocols::sumcheck::equal_n_vars_check, witness::MultilinearWitness};
+
+#[derive(Clone)]
+pub struct BaseExponentWitness<'a, P: PackedField> {
+	pub exponent: Vec<MultilinearWitness<'a, P>>,
+	pub single_bit_output_layers_data: Vec<MultilinearWitness<'a, P>>,
+	pub base: Option<MultilinearWitness<'a, P>>,
```
jimpo-ulvt, line 23: When this is `None` what is the base? The generator? This needs documentation.

crates/core/src/protocols/gkr_exponentiation/common.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2024-2025 Irreducible Inc.
+
+use binius_field::Field;
+
+use crate::protocols::gkr_gpa::LayerClaim;
+
+#[derive(Clone)]
+pub struct ExponentiationClaim<F: Field> {
```
jimpo-ulvt, line 8: Docstring

jimpo-ulvt (commented):
This is very complex, but understandable given the existing infrastructure for batched sumcheck and GKR. I can't recommend a better approach right now.

This PR does need much better documentation. Please make sure that all exported (public) structs, methods, and functions are documented.

crates/core/src/protocols/gkr_exponentiation/verifiers.rs
```diff
@@ -0,0 +1,239 @@
...
+
+	fn layer_composite_sum_claim(
+		&self,
+		layer_no: usize,
+		composite_claims_n_multilinears: usize,
+		multilinears_index: usize,
+		eq_ind_index: usize,
+	) -> Result<Option<CompositeSumClaim<F, VerifierExponentiationComposition<F>>>, Error>;
+
+	// return a tuple of the number of multilinears and the sumcheck claims used by this verifier for the current layer
```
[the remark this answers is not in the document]
djadjka: updated

crates/core/src/protocols/gkr_exponentiation/common.rs
```diff
@@ -0,0 +1,27 @@
...
+	fn from(value: ExponentiationClaim<F>) -> Self {
+		Self {
+			eval: value.eval,
+			eval_point: value.eval_point,
+		}
+	}
+}
+
+pub struct BaseExponentReductionOutput<F: Field> {
+	pub eval_claims_on_exponent_bit_columns: Vec<Vec<LayerClaim<F>>>,
```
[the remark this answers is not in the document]
djadjka: Yes, you're right, updated

crates/core/src/protocols/gkr_exp/batch_verify.rs
```diff
@@ -0,0 +1,260 @@
...
+	transcript::VerifierTranscript,
+	transparent::eq_ind::EqIndPartialEval,
+};
+
+/// Verify a batched GKR exponentiation protocol execution.
+///
+/// Since the protocol is verify bit-by-bit on each layer, we can batch verifiers starting from the first layer.
+///
+/// This protocol groups consecutive verifiers by their `eval_point` and reduces them to sumcheck provers on each layer.
+/// The sumcheck output, in turn, is reduce to LayerClaim's that uses for Evalcheck and internal prover [ExpClaim] for next layer.
```
jimpo-ulvt, line 25: Keep comments within 100 char line limit please

crates/core/src/protocols/gkr_exp/verifiers.rs
```diff
@@ -0,0 +1,290 @@
...
+	/// update prover internal [ExpClaim] and return the exponent bit [LayerClaim]
+	fn finish_layer(
+		&mut self,
+		layer_no: usize,
+		multilinear_evals: &[F],
+		r: &[F],
+	) -> Vec<LayerClaim<F>>;
+}
+
+pub struct GeneratorExpVerifier<F: Field, FBase>(ExpClaim<F>, PhantomData<FBase>);
```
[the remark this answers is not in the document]
djadjka: I put them in variables to make it more clear when reading the code

jimpo-ulvt (dismissed):

onesk (commented):
Lgtm.

onesk (approved):

jimpo-ulvt (dismissed):

onesk (approved):

PR #44, opened 2025-02-24
Add 8b-byte sliced fields that can be used as base for byte-sliced extensions

GraDKh:
Add 8b packed fields with byte-sliced memory layout equivalent to the byte-sliced fields we have.
Note that even though there are packed fields with the same number of scalar elements  their layout are different, that's why we use the following notation `NxMx8b`, where M is the number of bytes in SIMD register, N is the number of M-byte values in the memory representation.


crates/field/src/arch/portable/byte_sliced/packed_byte_sliced.rs
```diff
@@ -7,22 +7,22 @@ use std::{
...
-	PackedAESBinaryField16x8b, PackedAESBinaryField64x8b, PackedField,
+	PackedAESBinaryField16x8b, PackedAESBinaryField64x8b, PackedExtension, PackedField,
 };
 
-/// Represents 32 AES Tower Field elements in byte-sliced form backed by Packed 32x8b AES fields.
-///
-/// This allows us to multiply 32 128b values in parallel using an efficient tower
-/// multiplication circuit on GFNI machines, since multiplication of two 32x8b field elements is
-/// handled in one instruction.
+/// Represents AES Tower Field elements in byte-sliced form backed by Packed Nx8b AES fields
```
jimpo-ulvt, line 23: Please make the docstring summary sentence shorter. See docstring guidelines [here](https://github.com/IrreducibleOSS/binius/blob/main/CONTRIBUTING.md#documentation).

jimpo-ulvt: I'll merge anyway because this macro is not exported, but please follow the docstring guidelines going forward.

jimpo-ulvt (dismissed):

crates/field/src/arch/portable/byte_sliced/packed_byte_sliced.rs
GraDKh: Applied

jimpo-ulvt (approved):

PR #46, opened 2025-02-26
Implement linear transformations for byte-sliced fields

GraDKh:
Implement linear transformation for byte sliced fields. This functionality is necessary for the efficient `Vision32` implementation with byte-sliced fields. And even without it is much more efficient than linear transformations for the traditional packed fields as with one instruction we have 16 useful results of 8x8 matrix multiplication by 8b vector instead of 2.


djadjka (approved):
LGTM

jimpo-ulvt (approved):

PR #47, opened 2025-02-26
[ntt] Refactor AdditiveNTT to support more choices of NTT domain

jimpo-ulvt:
This is a refactor that

* Changes `AdditiveNTT` to be parameterized by `F: Field` instead of `P: PackedField`
* Exposes the NTT domain as a `BinarySubspace`

These changes are intended to support future changes that enable NTTs on FRI-compatible additive subspaces. The way NTT domains work currently, additive NTTs created for differently sized domains over the same basis elements are not FRI-compatible.


onesk (approved):
Lgtm.

PR #49, opened 2025-02-26
Disable benchmarking alternative arithmetic strategies by default

GraDKh:


GraDKh:
> Could we just put these in a different file and get the same benefit by just using `cargo bench --bench`?

That would be technically possible, but I'd rather go with a feature. Because if we split files we will have:
1. 5 more benchmarks files in the folder, each of the is a separate target and will slow the compilation down. Also it will bloat the benchmarks folder with similar files.
2. There are tiny commin pieces about `main` strategy which will be just copy-pasted.

I think its OK to have a feature, because that alternative strategies need to be measured rarely and I locally usually just comment them out.

Anex007 (approved):
Looks good to me

jimpo-ulvt (approved):

PR #55, opened 2025-02-28
[m3] Introduce new builder API

bergkvist:
Some of the features compared to `binius_circuits`:
- Decoupled witness generation from oracle defintions
- Explicit tables
- Typed columns with tower level and pack factor
- ConstraintSystem without table sizes that can be "compiled" into a ConstraintSystem
- No more "global shifts" like the one used in the old fibonacci example. The old fibonacci example follows more of an "AIR"-pattern, using global shifts to describe transition constraints (which we are moving away from). With M3/channels, you can think of each row as an entry in a multiset (where order does not matter), rather than an entry in a list/array (ordered). Channel pushes/pulls then achieve the same functionality as transition constraints in a more flexible and less constraining way.


jimpo-ulvt:
Awesome! I'm reviewing and adding documentation. In the meantime, can you split out the changes to `binius_math` and open a separate PR for that?

crates/m3/src/builder/column.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// The column has entries that are elements of `F`. In practice, the fields used will always be
+/// from the canonical tower (B1, B8, B16, B32, B64, B128). The second constant represents how many
+/// elements are packed vertically into a single logical row. For example, a column of type
+/// `Col<B1, 5>` will have 2^5 = 32 elements of `B1` packed into a single row.
+#[derive(Debug, Clone, Copy)]
+pub struct Col<F: TowerField, const V: usize = 0> {
+	// TODO: Maybe V should be powers of 2 instead of logarithmic
+	pub table_id: TableId,
+	pub partition: usize,
```
jimpo-ulvt, line 24: Is this always equal to the log packing factor from the `ColumnShape`? Does it need to be part of the ID?

bergkvist: It is not equal to the log packing factor, but rather the index of the partition within the table (which like column index depends on order of insertion)

jimpo-ulvt: But the partitions are grouped by packing factor or they might not be?

crates/m3/src/builder/column.rs
```diff
@@ -0,0 +1,113 @@
...
+/// A typed identifier for a column in a table.
+///
+/// The column has entries that are elements of `F`. In practice, the fields used will always be
+/// from the canonical tower (B1, B8, B16, B32, B64, B128). The second constant represents how many
+/// elements are packed vertically into a single logical row. For example, a column of type
+/// `Col<B1, 5>` will have 2^5 = 32 elements of `B1` packed into a single row.
+#[derive(Debug, Clone, Copy)]
+pub struct Col<F: TowerField, const V: usize = 0> {
+	// TODO: Maybe V should be powers of 2 instead of logarithmic
+	pub table_id: TableId,
```
jimpo-ulvt, line 23: Now that there are three fields, maybe just make the struct have `pub id: ColumnId, _marker: PhantomData<F>`

jimpo-ulvt (approved):
Going to merge an initial version and handle the rest as folllow-ups

jimpo-ulvt (approved):
We'll merge and address TODOs as follow-ups

onesk (approved):

PR #56, opened 2025-02-28
chore: Blake3 circuit usage example

storojs72:
This is a small follow-up PR with Blake3 example. It also includes [advised](https://github.com/IrreducibleOSS/binius/pull/38#discussion_r1973585875) refactoring of G function output (tuple -> array)


crates/circuits/src/blake3.rs
```diff
@@ -298,10 +298,10 @@ mod tests {
 
 		if let Some(witness) = builder.witness() {
 			(
-				witness.get::<F1>(output.0).unwrap().as_slice::<u32>(),
-				witness.get::<F1>(output.1).unwrap().as_slice::<u32>(),
-				witness.get::<F1>(output.2).unwrap().as_slice::<u32>(),
-				witness.get::<F1>(output.3).unwrap().as_slice::<u32>(),
+				witness.get::<F1>(output[0]).unwrap().as_slice::<u32>(),
```
onesk, line 301: Please refrain from using `.unwrap()` in library code. 

Also I would rewrite this as:
```
let [a, b, c, d] = output.map(|id| witness.get::<F1>(id).expect("oracle is present").as_slice::<u32>());
(a,b,c,d).par_iter()..
```

onesk (approved):
Nice, but have one nit. Please have a look.

crates/circuits/src/blake3.rs
storojs72: Though this is indeed included in the library, actually this is a unit-test code, so panicking here is not critical I think

PR #57, opened 2025-02-28
[piop] Parallelize data copying when merging multilinears

### LINKED ISSUE #51: [piop] Parallelize data copying when merging multilinears
See [this TODO comment](https://github.com/IrreducibleOSS/binius/blob/c035caef0bda2d5e187c362cb79ea776f0f1cc21/crates/core/src/piop/prove.rs#L53). Parallelize the data copy using the `binius_maybe_rayon` crate.

TomWambsgans:
Attempt to fix #51

- Should the 2nd copy phase (with n_packed_vars < P::log_width) also be parallelized ? (less important because it only copies small polynomials)
- There is a duplicated function `merge_multilins` [here](https://github.com/IrreducibleOSS/binius/blob/72ba99e63dfeb29227309e4c89d72b6d0ef6a5c5/crates/core/benches/poly_commit.rs#L27C4-L27C19) , is it intended ?
- Additional remark: The 2nd copy phase is never entered during tests

### DIFF
--- a/crates/core/src/piop/prove.rs
+++ b/crates/core/src/piop/prove.rs
@@ -51,31 +51,35 @@ where
 	P: PackedField,
 	M: MultilinearPoly<P>,
 {
-	// TODO: Parallelize the data copying
-	let mut packed_offset = 0;
 	let mut mle_iter = multilins.iter().rev();
 
 	// First copy all the polynomials where the number of elements is a multiple of the packing
 	// width.
 	let get_n_packed_vars = |mle: &M| mle.n_vars() - mle.log_extension_degree();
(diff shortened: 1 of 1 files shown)


jimpo-ulvt:
@Anex007 Why is `merge_multilins` duplicated in the benchmark file? Is it because you wanted to avoid the RS code construction in the benchmark?

jimpo-ulvt (approved):
Very nice!!

No need to parallelize the second loop, it shouldn't be very much work even if it gets executed.

Anex007:
> @Anex007 Why is `merge_multilins` duplicated in the benchmark file? Is it because you wanted to avoid the RS code construction in the benchmark?

Yep, I copied that code to avoid the creation of RS code in the benchmark

PR #61, opened 2025-03-01
New MultilinearPolyVariant for multilinear extensions of composites

### LINKED ISSUE #40: New MultilinearPolyVariant for multilinear extensions of composites
`binius_core::oracle::MultilinearPolyVariant` defines the different types of virtual multilinear oracles. One of them, for example, is `LinearCombination`, which is defined by a vector of pairs of other columns and associated scalar coefficients. Since it is a linear combination of multilinear polynomials, the derived polynomial is also multilinear and can be easily evaluated by a verifier, given oracle access to the underlying oracles.

We can define a multivariate polynomial as a composite of other multilinear oracles with a higher-degree composition. For example, given two n-variate oracles, $p_0$ and $p_1$, we can define the composite $f = p_0 * p_1$. Unfortunately, this would not be multilinear and thus could not be a `MultilinearPolyVariant`. On the other hand, the multilinear extension of a composite is well-defined. In fact, this satisfies our definition of a virtual polynomial oracle because there is a succinct reduction for evaluation: the sumcheck protocol. The verifier can check the evaluation of the MLE of a composite polynomial at a given point by a sumcheck reduction of the composite polynomial multiplied by a partially evaluated eq indicator.

This task then requires a few items:

1. Define the new `MultilinearPolyVariant`
2. Handle the evaluation of the new oracle type in the `binius_core::protocols::evalcheck` protocol.
3. Expose the new oracle variant through the `binius_circuits::ConstraintSystemBuilder` interface.

This feature would enable the functionality of data-parallel GKR, programmable with the standard Binius constraint system interface.

This task is moderately difficult, but possible for external contributors.

TomWambsgans:
Attempt to fix #40


ewile:
NOT ready for merge: Even though the new eval_check pass, it doesnt work on a [real project](https://github.com/TomWambsgans/conservative-xmss-aggregation/tree/binius-composite) (cargo run panics on `Evalcheck(Evalcheck(Verification(IncorrectEvaluation("Composition: merkle_table::pre_hash_main"))))`). Investigation in progress

TomWambsgans:
NOT ready for merge: even though the new evalcheck pass, it doesnt work end to end on [this project](https://github.com/TomWambsgans/conservative-xmss-aggregation/tree/binius-composite) (cargo run panics on `Evalcheck(Evalcheck(Verification(IncorrectEvaluation("Composition: merkle_table::pre_hash_main"))))`) . Investigation in progress

jimpo-ulvt:
Nice job!! I'll review when it passes for your project. Please let me know if you need help investigating/debugging.

fkondej (approved):

TomWambsgans:
The error appears in [`make_new_eval_claim`](https://github.com/IrreducibleOSS/binius/blob/26a8394bb064ea8b4a520caaf21e08ab3e9430fe/crates/core/src/protocols/evalcheck/prove.rs#L609) .

Contrary to all the other variants,`MultilinearPolyVariant::Composite` is not multilinear, and thus we cannot rely on `witness_poly` (`impl MultilinearPoly`, derived from the evaluations on the hypercube in the witness provided by the prover if I understand correctly) to compute an evaluation.

To evaluate a composition of multilinears on a random point P, one should instead evaluate each inner multilinear M1, M2, … on P, and then evaluate the outer composition (`ArithCircuitPoly`) on these inner evaluations M1(P), M2(P), … For the oracles of type `Composite`, the witness provided by the prover does not seem useful (in this case, should a user be able to fill it even though these values are never used?)

Problem: what if a user uses some oracles of type `MultilinearPolyVariant::Composite` to build another oracle (example: a linear combination of `Composite`). The resulting oracles would not be multivariate and the same problem in `make_new_eval_claim` appears. I see 2 solutions:
- Forbid the use of `Composite` oracle as source to construct other oracles
- Make the user (prover) fill the witness only for committed oracles. All the other oracles can be derived during the proof generation (linear combinations, shifts, compositions…), but I feel this requires a big refactor in the code.

Am I missing something simpler ?

TomWambsgans:
My mistake is that I've built a "true" composition of multilinears, instead of the MLE of the evaluations of a composition of multlinears on the hypercube. Should be fixed in the coming days

jimpo-ulvt:
> My mistake is that I've built a "true" composition of multilinears, instead of the MLE of the evaluations of a composition of multlinears on the hypercube. Should be fixed in the coming days

Yes, exactly. Let me know if you have questions about how to implement that.

TomWambsgans:
It's already better, now it's working on the [other project](https://github.com/TomWambsgans/conservative-xmss-aggregation/tree/binius-composite). However, there are still some buggy configurations (c.f. my latest commit). I've still got some work to do

crates/core/src/constraint_system/validate.rs
```diff
@@ -244,6 +244,21 @@ where
 				});
 			}
 		}
+		MultilinearPolyVariant::Composite(composition) => {
+			let uncombined_polys = composition
```
jimpo-ulvt, line 248: Might be helpful to use the `MultilinearComposite` struct here to reuse code.

crates/core/src/oracle/multilinear.rs
```diff
@@ -722,6 +773,55 @@ impl<F: TowerField> LinearCombination<F> {
 	}
 }
 
+/// MLE of a multivariate polynomial evaluated on multilinear oracles.
+///
+/// i.e. the MLE of the evaluations of C(M1, M2, ..., Mn) on {0, 1}^μ where:
```
jimpo-ulvt, line 778: tip: you can use LaTeX syntax in docstrings and it gets rendered correctly on https://docs.binius.xyz/binius_core/ because we've included the KaTeX JavaScript plugin there. So this will read better as:

```
/// i.e. the MLE of the evaluations of $C(M_1, M_2, \ldots, M_n)$ on $\{0, 1\}^\mu$ where:
```

jimpo-ulvt (commented):
Nice!

@djadjka please review evalcheck code
@bergkvist please review oracle serialization code

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -123,7 +124,7 @@ enum CircuitStep<F: Field> {
 /// and the object representing different polnomials can be stored in a homogeneous collection.
 #[derive(Debug, Clone)]
 pub struct ArithCircuitPoly<F: Field> {
-	expr: ArithExpr<F>,
+	pub(crate) expr: ArithExpr<F>,
```
bergkvist, line 127: Instead of making this field public, I recommend adding a method `pub fn expr(&self) -> &ArithExpr<F>` since that makes it easier to reason about potential mutation of the field. You can do this slightly less verbosely using the `Getters` derive macro.

bergkvist:
Serialization logic looks good to me

TomWambsgans:
Thank you for your feedback. The suggestions have been committed. Now I need to debug the failing example. I don't like my naming (`MultilinearComposition` is confusing with`MultilinearComposite`), I will polish things once all bugs are fixed.

jimpo-ulvt:
> Thank you for your feedback. The suggestions have been committed. Now I need to debug the failing example. I don't like my naming (`MultilinearComposition` is confusing with`MultilinearComposite`), I will polish things once all bugs are fixed.

We use "composition" to refer to `C(X, Y, Z)` and "composite" to refer to `C(F_1(X_1, ..., X_N), F_2(X_1, ..., X_N), F_3(X_1, ..., X_N))`.

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -383,6 +386,14 @@ where
 				self.incomplete_proof_claims
 					.insert(multilinear_id, eval_point, evalcheck_claim);
 			}
+
+			MultilinearPolyVariant::Composite(_) => {
```
djadjka, line 390: Perhaps it would make sense to move it next to packed/shifted.

djadjka (commented):
Evalcheck looks good to me

TomWambsgans:
I have fixed the bug. I have written a [complex example](https://github.com/IrreducibleOSS/binius/pull/61/files#diff-85a24fd71c6355c0741a227c2d95209bda0e674046712ae9351b2affe348583f), which works fine. I changed some naming (`MultilinearPolyVariant::Composite(CompositeMLE)`. `CompositeMLE` to avoid the confusion with `MultilinearComposite`). 

I think (and hope) everything works now, but I still need to improve the code quality, especially in `subclaims.rs`:
I have reused the structure `ProjectedBivariateMeta` (previously used for shifted + packed oracles, which are the oracles whose evalcheck is reduced to a sumcheck). The new “composite” oracle also reduces an evalcheck to a sumcheck, but the expression in the sumcheck is not bivariate as in shift / pack. Instead, it is equal to the circuit of the composite oracle, multiplied by “eq” MLE. As a result, the naming in `subclaims.rs` is confusing currently because everything has been planned for a bivariate sumcheck.

I don’t know if I am supposed to keep the end-to-end example I have written, and if yes, in which file / folder to put it ? (Currently it is in `/examples`)

TomWambsgans:
Other remark: in [`process_sumcheck`](https://github.com/IrreducibleOSS/binius/blob/0a6bdc7c3897c0684323ce7331a1550b7e4167f1/crates/core/src/protocols/evalcheck/prove.rs#L236), for the composite oracles, I compute the witness for the eq MLE (c.f. `backend` which is passed as parameter, [here](https://github.com/IrreducibleOSS/binius/blob/0a6bdc7c3897c0684323ce7331a1550b7e4167f1/crates/core/src/protocols/evalcheck/prove.rs#L629)). This is bad because it is not parallelized. I should probably compute it in the previous function [calculate_projected_mles](https://github.com/IrreducibleOSS/binius/blob/0a6bdc7c3897c0684323ce7331a1550b7e4167f1/crates/core/src/protocols/evalcheck/prove.rs#L223). Currently, for composite oracles, I believe `calculate_projected_mles` re-computes the evaluation of the composite oracle on the `eval_point`, which is useless. Instead, I think it should compute the eq MLE associated with the composite oracle.

djadjka:
We mostly use examples for circuits, so I think you can move it to the tests and just add ignore if it's slow.

As for your other comments, let me check them again and I will return to you soon

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -183,6 +208,44 @@ where
...
+	witness_index: &mut MultilinearExtensionIndex<U, F>,
+	constraint_builders: &mut Vec<ConstraintSetBuilder<F>>,
+	backend: &impl ComputationBackend,
+	projected: MultilinearExtension<PackedType<U, F>>,
+) -> Result<(), Error>
+where
+	U: UnderlierType + PackScalar<F>,
+	F: TowerField,
+{
+	process_projected_bivariate_witness(
```
djadjka, line 226: You can just add poly to the witness without using process_projected_bivariate_witness, and you're right about parallelization. However, I think we can take data from memoized_queries since it's also tensor products.

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -183,6 +208,44 @@ where
...
+	)?;
+
+	let mut oracle_ids = comp.inner().clone();
+	oracle_ids.push(meta.multiplier_id);
+
+	if comp.n_vars() > constraint_builders.len() {
+		constraint_builders.resize_with(comp.n_vars(), || ConstraintSetBuilder::new());
+	}
+	// Var(comp.n_polys()) corresponds to the eq MLE (meta.multiplier_id)
+	let expr = <_ as CompositionPoly<F>>::expression(comp.c()) * ArithExpr::Var(comp.n_polys());
```
djadjka, line 245: you can use `add_composite_sumcheck_to_constraints` here

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -126,7 +129,20 @@ pub fn packed_sumcheck_meta<F: TowerField>(
...
 
-pub fn add_bivariate_sumcheck_to_constraints<F: Field>(
+pub fn composite_sumcheck_meta<F: TowerField>(
+	oracles: &mut MultilinearOracleSet<F>,
+	id: OracleId,
+	eval_point: &[F],
+) -> Result<ProjectedBivariateMeta, Error> {
+	Ok(ProjectedBivariateMeta {
+		multiplier_id: oracles.add_transparent(EqIndPartialEval::new(eval_point.to_vec()))?,
+		inner_id: id,
```
djadjka, line 139: Maybe it makes sense to do inner_id as an Option too; it would allow you to filter projected and composite in calculate_projected_mles.

djadjka (commented):
I agree that using ProjectedBivariate is not the best approach, but using separate structures, IMHO, would unnecessarily complicate the code. In general, it can still be named Bivariate since you use $eq\\_ind \cdot composition$.

jimpo-ulvt:
This looks great! Can you please rebase this onto main and resolve the conflicts? It should be good to merge after that.

crates/core/src/protocols/evalcheck/subclaims.rs
TomWambsgans: I made `inner_id` an Option, and I made `calculate_projected_mles` compute the eq MLE for composite oracles. It is better, thanks for the suggestion. I still find the distinction between shifted / packed / composite oracles hard to follow in `subclaims.rs`, but maybe it is ok like this ? I could try to use an enum for `ProjectedBivariateMeta` ?

TomWambsgans: (I renamed `calculate_projected_mles` to  `calculate_meta_mles` because for the composite oracles it is not really a projection, simply an eq MLE). If you have a better name I am interested.

TomWambsgans:
> This looks great! Can you please rebase this onto main and resolve the conflicts? It should be good to merge after that.

Done

TomWambsgans:
I think some additional work is required to improve the code quality in `subclaims.rs`, because the struct [`ProjectedBivariateMeta`](https://github.com/IrreducibleOSS/binius/blob/e2481a353f5b708dff3d7113bc65e9dfa4eb4d85/crates/core/src/protocols/evalcheck/subclaims.rs#L235) can now represent packed / shifted / composite, and this is (at least for me) a bit confusing. However this could be improved in the future after merging this PR, depends on how readable you find it currently

djadjka:
@TomWambsgans Sorry, I want to review your update before merging.
I will try to do it tomorrow morning.
thank you!

TomWambsgans:
> @TomWambsgans Sorry, I want to review your update before merging. I will try to do it tomorrow morning. thank you!

Yes, of course, especially since I'm not completely satisfied with `subclaims.rs`.

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -183,9 +207,33 @@ where
...
+	eval: F,
+	witness_index: &mut MultilinearExtensionIndex<U, F>,
+	constraint_builders: &mut Vec<ConstraintSetBuilder<F>>,
+	eq_mle: MultilinearExtension<PackedType<U, F>>,
+) -> Result<(), Error>
+where
+	U: UnderlierType + PackScalar<F>,
+	F: TowerField,
+{
+	if !witness_index.has(meta.multiplier_id) {
```
djadjka, line 223: Unfortunately, this block will always execute because you add a new multiplier_id for every composite case

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -183,9 +207,33 @@ where
...
+	}
+
+	add_composite_sumcheck_to_constraints(meta, constraint_builders, comp, eval);
+	Ok(())
+}
+
 #[derive(Clone, Copy)]
 pub struct ProjectedBivariateMeta {
-	inner_id: OracleId,
+	inner_id: Option<OracleId>, // Some if shifted / packed, None if composite
```
djadjka, line 236: if you don't mind, could you please add this comment as a Rust doc comment?
Thank you!

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -285,34 +338,44 @@ where
...
-				.map_err(Error::from)
-		})
-		.collect::<Result<Vec<_>, Error>>()?;
-
-	// Memoize queries for calculate_projected_mle
+	let mut queries_to_memoize = Vec::new();
 	for (meta, claim) in metas.iter().zip(projected_bivariate_claims) {
-		let eval_point = &claim.eval_point[meta.projected_n_vars..];
-		memoized_queries.full_query(eval_point, backend)?;
+		let point = if meta.inner_id.is_some() {
```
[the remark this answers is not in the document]
TomWambsgans: I think the line 172 only concerns the eval_points in `claims_without_evals` (which come from linear combinations and zero padded oracles) ? if I am right, we need to memoize again for the eq MLE coming from composite oracles (which I do in the new function `fill_eq_witness_for_composites` I have written following your suggestion. Please correct me if I am wrong

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -273,48 +301,102 @@ where
...
+
+	// composite eval_point -> eq MLE witness
+	let eq_mles: HashMap<&[F], MultilinearWitness<'_, _>> = composite_eval_points
+		.iter()
+		.map(|eval_point| {
+			let mle = MLEDirectAdapter::from(MultilinearExtension::from_values(
+				memoized_queries
+					.full_query(eval_point, backend)?
+					.expansion()
+					.to_vec(),
```
djadjka, line 385: Unfortunately, there is still unnecessary cloning for the same val_points going on here. `clone -> to Arc -> drop Arc`

djadjka: I would suggest doing it this way.

```rust
let dedup_eval_points = metas
		.iter()
		.zip(projected_bivariate_claims)
		.filter(|(meta, _)| meta.inner_id.is_none())
		.map(|(_, claim)| claim.eval_point.as_ref())
		.collect::<HashSet<_>>();

	let eq_indicators = dedup_eval_points
		.into_iter()
		.map(|eval_point| {
			let mle = MLEDirectAdapter::from(MultilinearExtension::from_values(
				memoized_queries
					.full_query(eval_point, backend)?
					.expansion()
					.to_vec(),
			)?)
			.upcast_arc_dyn();
			Ok((eval_point, mle))
		})
		.collect::<Result<HashMap<_, _>, Error>>()?;

	for (meta, claim) in metas
		.iter()
		.zip(projected_bivariate_claims)
		.filter(|(meta, _)| meta.inner_id.is_none())
	{
		let eq_ind = eq_indicators
			.get(claim.eval_point.as_ref())
			.expect("was added above");

		witness_index.update_multilin_poly(vec![(meta.multiplier_id, eq_ind_witness.clone())])?;
	}
```

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -324,60 +301,102 @@ where
...
+{
+	let (composite_metas, composite_eval_points): (Vec<&ProjectedBivariateMeta>, Vec<&[F]>) =
+		Iterator::unzip(
+			metas
+				.iter()
+				.zip(projected_bivariate_claims)
+				.filter(|(meta, _)| meta.inner_id.is_none())
+				.map(|(meta, claim)| (meta, claim.eval_point.as_ref())),
+		);
+	memoized_queries.memoize_query_par(&composite_eval_points, backend)?;
```
djadjka, line 375: And we can drop memoized_queries.memoize_query_par because full_query memorizes it. If I'm not mistaken, this will only happen for the evalcheck challenge, because you're right, and we don't add it to the memoization in the prove.

djadjka (commented):
Here’s a small suggestion for improvement =)

crates/core/src/protocols/evalcheck/subclaims.rs
TomWambsgans: > Unfortunately, there is still unnecessary cloning for the same val_points going on here. `clone -> to Arc -> drop Arc`
  
Yes indeed, thanks for catching it. Your code looks good to me, I will use it. I just added a `memoized_queries
		.memoize_query_par(&dedup_eval_points, backend)` instead of the sequential  `full_query`.

crates/core/src/protocols/evalcheck/subclaims.rs
TomWambsgans: Why not doing it in parrallel with `memoize_query_par`  (instead of consecutive `full_query`) ? I am not sure to understand

djadjka: Oh, sorry, I have yogurt for brains. I switched to your PR from another protocol where there was only one eval_point.
Yes, here theoretically there could be several points, and your code is better.

jimpo-ulvt (approved):

PR #63, opened 2025-03-03
sumcheck: `equal_n_vars_check` earlier return when m empty

tcoratger:


crates/core/src/protocols/sumcheck/common.rs
```diff
@@ -265,16 +265,15 @@ where
...
-	let n_vars = multilinears
-		.first()
-		.map(|multilinear| multilinear.n_vars())
-		.unwrap_or_default();
-	for multilinear in multilinears {
-		if multilinear.n_vars() != n_vars {
-			bail!(Error::NumberOfVariablesMismatch);
-		}
+	let Some(n_vars) = multilinears.first().map(|m| m.n_vars()) else {
+		return Ok(0);
```
tcoratger, line 269: I think that here if `multilinears` is empty we can directly return 0 earlier as the following loop will do nothing in this case

jimpo-ulvt (commented):
I don't see the value of this? This isn't a hot/performance-critical routine.

crates/core/src/protocols/sumcheck/common.rs
tcoratger: Yes this seems just a bit easier to read for me so that if multilinears is empty then we just return 0. But this is personal, feel free to close if you judge it's useless

PR #76, opened 2025-03-03
[ci]: Setup rust as pre run & allow deadcode

milosbackonja:


crates/field/src/arch/portable/packed.rs
```diff
@@ -458,6 +458,7 @@ pub(crate) use impl_ops_for_zero_height;
 
 /// Multiply `PT1` values by upcasting to wider `PT2` type with the same scalar.
 /// This is useful for the cases when SIMD multiplication is faster.
+#[allow(dead_code)]
```
jimpo-ulvt, line 461: @GraDKh Does this look right?

GraDKh: This is valid for the platforms where we don't have SIMD at all, in this case this function is not used

GraDKh:
It's weird, but when I'm running the `clippy` command locally it doesn't pass:
```
cargo clippy --all --all-features --tests --benches --examples -- -D warnings
```
This is actually because we have a feature `stable_only` with which we currently can't compile all the crate. I suppose we shouldn't run clippy with `--all-features` and pick some reasonable subset

jimpo-ulvt (approved):

fkondej (approved):

PR #83, opened 2025-03-04
[m3] Add TableBuilder with namespacing

bergkvist:
Moves `Table::add_*` methods to `TableBuilder::add_*`, which also supports namespacing.

To add a namespace before passing the TableBuilder to a gadget, you can do something like:
```rs
let inner_table = Table::new(..)
let table = TableBuilder::new(&mut inner_table)
U32Add::new(
  &mut table.with_namespace("u32add"),
  ..
)
```

### DIFF
--- a/crates/m3/src/builder/table.rs
+++ b/crates/m3/src/builder/table.rs
@@ -189,7 +189,7 @@ pub struct Table<F: TowerField = B128> {
 	pub id: TableId,
 	pub name: String,
 	pub columns: Vec<ColumnInfo<F>>,
-	pub partitions: SparseIndex<TablePartition<F>>,
+	pub(super) partitions: SparseIndex<TablePartition<F>>,
 	/// This indicates whether a table is fixed for constraint system or part of the dynamic trace.
 	///
 	/// Fixed tables are either entirely transparent or committed during a preprocessing step that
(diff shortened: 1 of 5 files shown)


crates/m3/tests/collatz.rs
```diff
@@ -218,7 +218,7 @@ mod arithmetization {
 
 			// Input times 3 + 1
 			let triple_plus_one = U32Add::new(
-				table,
+				&mut table.with_namespace("u32add"),
```
jimpo-ulvt, line 221: Please change the namespace to `"triple_plus_one"`. It's more useful to label it with it's function in the caller than what the gadget is exactly (which is known to the gadget already).

crates/m3/src/builder/table.rs
```diff
@@ -247,23 +148,157 @@ impl<F: TowerField> Table<F> {
...
+	/// occurs before any statements are proven.
+	pub is_fixed: bool,
+}
+
+/// A table partition describes a part of a table where everything has the same pack factor (as well as height)
+/// Tower level does not need to be the same.
+///
+/// Zerocheck constraints can only be defined within table partitions.
+#[derive(Debug)]
+pub struct TablePartition<F: TowerField = B128> {
```
jimpo-ulvt, line 205: Not important in this PR, but I wonder if we can make this private or at least just `pub(super)` rather than exposing partitions publicly.

jimpo-ulvt (changes requested):

PR #90, opened 2025-03-05
[m3] Add prove and verify to collatz test

bergkvist:


jimpo-ulvt (approved):

jimpo-ulvt (changes requested):
Please refactor into two tests that share setup logic. One fast test for validation and one slower test for prove/verify.

jimpo-ulvt (approved):

PR #93, opened 2025-03-05
Multiplication based on gkr_exp

djadjka:
| Data Type | Method        | Prove Time | Verify Time |
|-----------|---------------|------------|-------------|
| u64       | GKR Exp       | 4.8s       | 56ms        |
| u32       | GKR Exp       | 2.74s      | 34.4ms      |
| u32       | Lookups       | 5.02s      | 37.25ms     |
| u8        | GKR Exp       | 1.14s      | 20.87ms     |
| u8        | Lookups       | 0.2s       | 9.31ms      |


crates/circuits/src/arithmetic/mul.rs
```diff
@@ -0,0 +1,238 @@
...
+	builder: &mut ConstraintSystemBuilder,
+	name: impl ToString,
+	xin_bits: Vec<OracleId>,
+	yin_bits: Vec<OracleId>,
+) -> Result<Vec<OracleId>, anyhow::Error> {
+	let name = name.to_string();
+
+	let log_rows = builder.log_rows([xin_bits.clone(), yin_bits.clone()].into_iter().flatten())?;
+
+	let xin_exp_result_id = builder.add_committed(
```
jimpo-ulvt, line 31: Leave comments with the mathematical values of these columns.

crates/core/src/constraint_system/mod.rs
```diff
@@ -29,6 +31,7 @@ pub struct ConstraintSystem<F: TowerField> {
 	pub table_constraints: Vec<ConstraintSet<F>>,
 	pub non_zero_oracle_ids: Vec<OracleId>,
 	pub flushes: Vec<Flush>,
+	pub mul: Vec<Mul>,
```
jimpo-ulvt, line 34: Instead of adding multiplications to the constraint system, add exponentiation relationships.

jimpo-ulvt (changes requested):

jimpo-ulvt:
@onesk Please review

crates/circuits/src/arithmetic/mul.rs
```diff
@@ -0,0 +1,264 @@
...
+		cout_low_exp_result_id,
+		ExpBase::Static(FExpBase::MULTIPLICATIVE_GENERATOR.into()),
+		FExpBase::TOWER_LEVEL,
+	);
+	builder.add_exp(
+		cout_high_bits.to_vec(),
+		cout_high_exp_result_id,
+		ExpBase::Static(
+			FExpBase::MULTIPLICATIVE_GENERATOR
+				.pow(1 << cout_low_bits.len())
```
onesk, line 154: As far as I can see you need `u128` exponents mostly to handle this, but:
 1) This is a power-of-two exponent, which admits a very neat repeated squarings implementation
 2) To me, it's hard to imagine `pow` taking `u128` being useful anywhere but there; interested to know your opinion

My gut feel would be to make an `exp_pow2` helper within multiplication code and roll back the `u64->u128` change in `PackedField::pow` which inflates this changeset.

crates/core/src/tower.rs
```diff
@@ -5,11 +5,15 @@
...
-	polyval::{AES_TO_POLYVAL_TRANSFORMATION, BINARY_TO_POLYVAL_TRANSFORMATION},
+	polyval::{
+		AES_TO_POLYVAL_TRANSFORMATION, BINARY_TO_POLYVAL_TRANSFORMATION,
+		POLYVAL_TO_AES_TRANSFORMARION,
+	},
 	underlier::UnderlierType,
 	AESTowerField128b, AESTowerField16b, AESTowerField32b, AESTowerField64b, AESTowerField8b,
 	BinaryField128b, BinaryField128bPolyval, BinaryField16b, BinaryField1b, BinaryField32b,
 	BinaryField64b, BinaryField8b, ExtensionField, PackedExtension, PackedField, TowerField,
+	POLYVAL_TO_BINARY_TRANSFORMATION,
```
onesk, line 16: nit: this looks stray, with every other transformation being imported from `binius_field::polyval` :)

onesk (commented):
A few comments; still in progress.

crates/core/src/constraint_system/exp.rs
```diff
@@ -0,0 +1,273 @@
...
+				.expect("poly contain packed_evals");
+
+			let packed_evals = PackedType::<U, Tower::B128>::cast_bases(packed_evals);
+
+			MultilinearExtension::new(exp_witness.n_vars(), packed_evals.to_vec())
+				.map(|mle| {
+					MLEEmbeddingAdapter::<PackedType<U, Tower::B1>, PackedType<U, Tower::FastB128>>::from(
+				mle,
+			)
+			.upcast_arc_dyn()
```
onesk, line 268: Can you use `specialize_arc_dyn` instead?

crates/core/src/constraint_system/exp.rs
```diff
@@ -0,0 +1,273 @@
...
+
+	MultilinearExtension::new(witness.n_vars(), fast_packed_evals)
+		.map(|mle| MLEDirectAdapter::from(mle).upcast_arc_dyn())
+		.map_err(Error::from)
+}
+
+type FastExponentWitnesses<'a, U, Tower> =
+	Vec<MultilinearWitness<'a, PackedType<U, FFastExt<Tower>>>>;
+
+fn get_fast_exponent_witnesses<'a, U, Tower>(
```
onesk, line 243: nit: please document this method, mentioning that it uses 1-bit embedding for the fast field (aka POLYVAL before we change to byteslicing).

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -216,6 +219,23 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
 			.committed(n_vars, tower_level)
 	}
 
+	/// Be careful, exp_result witness will be added during the proving stage,
+	/// so ensure the correct order is maintained.
```
onesk, line 223: Not sure what "correct order" is mentioned here; does it mean that witness generation code should be aware of the fact that exponent result is not initialized yet?

onesk (changes requested):

crates/circuits/src/builder/constraint_system.rs
djadjka: Updated to avoid this problem

crates/core/src/constraint_system/exp.rs
```diff
@@ -0,0 +1,342 @@
...
+};
+
+#[derive(Debug, Clone, SerializeBytes, DeserializeBytes)]
+pub struct Exp<F: Field> {
+	/// A vector of `OracleId`s representing the exponent in little-endian bit order
+	pub bits_ids: Vec<OracleId>,
+	pub base: ExpBase<F>,
+	pub exp_result_id: OracleId,
+	/// Specifies the field level in the tower where `base` is defined
+	pub base_tower_level: usize,
```
jimpo-ulvt, line 37: This seems like it should be a field on `ExpBase::Static` instead of this struct. `ExpBase::Dynamic` should already carry tower height info because we can query it from `MultilinearOracleSet`.

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -216,6 +219,28 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
...
 	}
 
+	/// Adds an exponentiation operation to the constraint system.
+	///
+	/// # Parameters
+	/// - `bits_ids`: A vector of `OracleId` representing the exponent in little-endian bit order.
+	/// - `exp_result_id`: The `OracleId` that holds the result of the exponentiation..
+	/// - `base`: The base value for exponentiation, represented as an [`ExpBase`].
+	/// - `base_tower_level`: Specifies the field level in the tower where `base` is defined
+	pub fn add_exp(
```
jimpo-ulvt, line 229: This can just take an `exp: Exp` argument. The struct has only `pub` fields -- the caller can create it themself.

crates/core/src/constraint_system/verify.rs
```diff
@@ -282,7 +301,8 @@ where
 		[non_zero_prodcheck_eval_claims, flush_eval_claims]
 			.concat()
 			.into_iter()
-			.chain(zerocheck_eval_claims),
+			.chain(zerocheck_eval_claims)
```
jimpo-ulvt, line 304: nit: you can just use the `itertools::chain!` macro for all of these:
```rust
chain!(
    non_zero_prodcheck_eval_claims,
    flush_eval_claims,
    zerocheck_eval_claims,
    exp_eval_claims,
)
```

crates/core/src/protocols/gkr_exp/witness.rs
```diff
@@ -1,177 +1,148 @@
...
+		.into_par_iter()
+		.map(|i| {
+			let base = match &base {
+				Base::Static(g) => *g,
+				Base::Dynamic(poly) => poly
+					.packed_evals()
+					.expect("base and exponent_bit_witness have the same width")[i],
+			};
+
+			// TODO: fix when MLEEmbeddingAdapter::packed_evals() works correct
```
jimpo-ulvt, line 93: What about it doesn't work? We should be able to check the `log_embedding_degree` and verify that it's actually a 1-bit field.

crates/core/src/protocols/gkr_exp/witness.rs
```diff
@@ -1,177 +1,148 @@
...
+			let exponent_bit = P::from_fn(|j| {
+				exponent_bit_witness
+					.evaluate_on_hypercube(P::WIDTH * i + j)
```
jimpo-ulvt, line 96: Maybe use `evaluate_on_hypercube_and_scale` here instead so the multiplication takes place before the upcast

crates/core/src/protocols/gkr_exp/witness.rs
```diff
@@ -1,177 +1,148 @@
...
+	Static(P),
+	Dynamic(MultilinearWitness<'a, P>),
 }
 
-fn evaluate_first_layer_output_packed<PBits, PBase>(
-	exponent_bit: &[PBits],
-	base: BaseEvals<PBase>,
-) -> Vec<PBase>
+fn evaluate_first_layer_output_packed<P>(
+	exponent_bit_witness: MultilinearWitness<P>,
```
jimpo-ulvt, line 74: That's too bad that we only know about the type-erased poly here.

crates/circuits/src/arithmetic/mul.rs
```diff
@@ -0,0 +1,262 @@
...
+
+		let mut cout_columns_u8 = cout_columns
+			.iter_mut()
+			.map(|column| column.as_mut_slice::<u8>())
+			.collect::<Vec<_>>();
+
+		numbers_to_columns(&result, &mut cout_columns_u8);
+	}
+
+	builder.assert_zero(
```
jimpo-ulvt, line 104: Leave a comment explaining this assertion.

crates/circuits/src/arithmetic/mul.rs
```diff
@@ -0,0 +1,262 @@
...
+				log_rows,
+				BinaryField1b::TOWER_LEVEL,
+			)
+		})
+		.collect::<Vec<_>>();
+
+	if let Some(witness) = builder.witness() {
+		let xin_columns = xin_bits
+			.iter()
+			.map(|&id| witness.get::<BinaryField1b>(id).map(|x| x.as_slice::<u8>()))
```
jimpo-ulvt, line 77: Is it actually better to convert to `u8` slices? Why not just keep them as 1-bit packed fields?

crates/circuits/src/builder/constraint_system.rs
djadjka: I don’t really like using Exp here. I split it into add_static_exp and add_dynamic_exp, as I think this is the prettiest way.

crates/core/src/protocols/gkr_exp/witness.rs
djadjka: I will revert support for PBase and PExpBase when we start using ByteSliced

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -216,6 +219,49 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
...
+	/// - `bits_ids`: A vector of `OracleId` representing the exponent in little-endian bit order.
+	/// - `exp_result_id`: The `OracleId` that holds the result of the exponentiation..
+	/// - `base`: The static base value.
+	/// - `base_tower_level`: Specifies the field level in the tower where `base` is defined
+	pub fn add_static_exp(
+		&mut self,
+		bits_ids: Vec<OracleId>,
+		exp_result_id: OracleId,
+		base: F,
+		base_tower_level: usize,
```
jimpo-ulvt, line 234: It would be better if this was generic over `FSub` where `F: ExtensionField<FSub>` and `base: FSub`. Then we could remove the `base_tower_level` parameter because it would be `FSub::TOWER_LEVEL`.

jimpo-ulvt (approved):

bergkvist (approved):

PR #98, opened 2025-03-05
Add log size parameter to AdditiveNTT methods

onesk:
Previously there was no way to specify an NTT forward/inverse transform narrower than a packing width, which incurred unneeded `PackedFieldIndexable` constraints in univariate skip. This PR adds `log_n` as the fourth parameter to `AdditiveNTT` methods and updates usage sites accordingly.

### DIFF
--- a/crates/ntt/src/multithreaded.rs
+++ b/crates/ntt/src/multithreaded.rs
@@ -115,7 +115,7 @@ fn forward_transform<F: BinaryField, P: PackedField<Scalar = F>>(
 					data,
 					coset,
 					log_batch_size,
-					P::LOG_WIDTH - log_batch_size,
+					log_n,
 				),
 			};
 		}
@@ -209,7 +209,7 @@ fn inverse_transform<F: BinaryField, P: PackedField<Scalar = F>>(
 					data,
 					coset,
 					log_batch_size,
-					log_n.saturating_sub(log_batch_size),
+					log_n,
 				),
 			};
 		}
--- a/crates/ntt/src/multithreaded.rs
+++ b/crates/ntt/src/multithreaded.rs
@@ -7,7 +7,7 @@ use binius_utils::rayon::get_log_max_threads;
 
 use super::{
 	error::Error,
-	single_threaded::{self, check_batch_transform_inputs, NTTParams},
+	single_threaded::{self, check_batch_transform_inputs_and_params},
 	strided_array::StridedArray2DViewMut,
 	twiddle::TwiddleAccess,
 	AdditiveNTT, SingleThreadedNTT,
@@ -63,6 +63,7 @@ where
 		data: &mut [P],
 		coset: u32,
 		log_batch_size: usize,
+		log_n: usize,
 	) -> Result<(), Error> {
 		forward_transform(
 			self.log_domain_size(),
@@ -70,6 +71,7 @@ where
 			data,
 			coset,
 			log_batch_size,
+			log_n,
 			self.log_max_threads,
 		)
 	}
@@ -79,6 +81,7 @@ where
 		data: &mut [P],
 		coset: u32,
 		log_batch_size: usize,
+		log_n: usize,
 	) -> Result<(), Error> {
 		inverse_transform(
 			self.log_domain_size(),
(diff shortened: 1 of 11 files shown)


crates/ntt/src/odd_interpolate.rs
```diff
@@ -67,7 +67,7 @@ impl<F: BinaryField> OddInterpolate<F> {
 		}
 
 		for (i, chunk) in data.chunks_exact_mut(1 << ell).enumerate() {
-			ntt.inverse_transform(chunk, i as u32, 0)?;
```
GraDKh, line 70: Isn't `chunk.len().ilog2() as usize)` equal to `ell` ?

GraDKh (commented):
Would be good to add tests for the case when `log_n` is less than `P::LOG_WIDTH`, as I understand it is something that didn't work before

crates/ntt/src/odd_interpolate.rs
onesk: That's very true, fixed. Also added the requested unit test, which actually caught a bug, thanks!

GraDKh (dismissed):

crates/ntt/src/single_threaded.rs
```diff
@@ -344,9 +313,13 @@ pub const fn check_batch_transform_inputs_and_params<PB: PackedField>(
 	}
 
 	let coset_bits = 32 - coset.leading_zeros() as usize;
-	if log_n + coset_bits > log_domain_size {
+
+	// The minimal domain size should cover at least two packed field elements
+	// for the implementation always be able to use PackedField::interleave.
+	let log_required_domain_size = (log_n + coset_bits).max(PB::LOG_WIDTH + 1);
```
jimpo-ulvt, line 319: I don't understand the comment and `PB::LOG_WIDTH` requirement. What about when the batch size is large?

crates/ntt/src/tests/ntt_tests.rs
```diff
@@ -22,31 +22,50 @@ use crate::{dynamic_dispatch::DynamicDispatchNTT, AdditiveNTT, SingleThreadedNTT
...
-		assert_eq!(data, &data_copy);
-		assert_eq!(data, &data_copy_2);
+	let log_n_range = if data.len() > 1 {
+		let single_log_n = data.len().ilog2() as usize + P::LOG_WIDTH - log_batch_size;
+		single_log_n..single_log_n + 1
+	} else {
+		0..P::LOG_WIDTH - log_batch_size
+	};
+
+	for log_n in log_n_range {
```
jimpo-ulvt, line 39: This test is confusing. Why is this looping? Can you just pass in the log_n parameter you want to test?

crates/ntt/src/single_threaded.rs
onesk: Tried to clarify the comment. Good catch about the batch size, fixed.

crates/ntt/src/tests/ntt_tests.rs
onesk: Hoisted the loop to the callsite. I've introduced sub- and multiple packed cases for `log_n` ranges mostly due to `BatchTooLarge` check.

crates/ntt/src/multithreaded.rs
```diff
@@ -110,16 +115,18 @@ fn forward_transform<F: BinaryField, P: PackedField<Scalar = F>>(
 					data,
 					coset,
 					log_batch_size,
-					P::LOG_WIDTH - log_batch_size,
+					log_n,
```
jimpo-ulvt, line 118: This whole clause seems unnecessary now actually

onesk: Can you please explain? I tried removing it and it fails. And I guess the intent was always to call into the corresponding `single_threaded` method which has sub-packing-width logic?

jimpo-ulvt: Oh, sorry, I got confused and thought this was `single_threaded.rs`. Nevermind.

jimpo-ulvt (approved):

PR #103, opened 2025-03-07
[m3] Add support for repeating constants

bergkvist:
Adding a repeating constant actually creates 2 oracles:
- transparent
- a repeating version of the transparent poly

We get around needing to fill the transparent witness, or even returning an id for it by having it alias the same underlying data as the repeating transparent (but with a different length value).

### DIFF
--- a/crates/m3/src/builder/column.rs
+++ b/crates/m3/src/builder/column.rs
@@ -1,8 +1,8 @@
 // Copyright 2025 Irreducible Inc.
 
-use std::marker::PhantomData;
+use std::{marker::PhantomData, sync::Arc};
 
-use binius_core::oracle::ShiftVariant;
+use binius_core::{oracle::ShiftVariant, polynomial::MultivariatePoly};
 use binius_field::{ExtensionField, TowerField};
 use binius_math::ArithExpr;
 
@@ -88,6 +88,8 @@ pub struct ColumnInfo<F: TowerField = B128> {
 	pub shape: ColumnShape,
 	/// Whether the column is constrained to be non-zero.
 	pub is_nonzero: bool,
+	/// Whether the column is transparent and only contains a single row.
+	pub is_single_row: bool,
 }
 
 /// The shape of each cell in a column.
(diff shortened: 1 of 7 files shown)


crates/m3/src/builder/column.rs
```diff
@@ -138,4 +140,10 @@ pub enum ColumnDef<F: TowerField = B128> {
 		cols: Vec<ColumnIndex>,
 		expr: ArithExpr<F>,
 	},
+	Transparent {
```
jimpo-ulvt, line 143: Can you consolidate these two into one variant? I don't think we need to express the non-repeating column as a separate column at the m3 layer -- just at the oracle layer.

crates/m3/src/builder/constraint_system.rs
```diff
@@ -320,18 +328,35 @@ fn add_oracle_for_column<F: TowerField>(
...
 				.iter()
 				.map(|&col_index| oracle_lookup[col_index])
 				.collect::<Vec<_>>();
-			addition.composite_mle(n_vars, inner_oracles, expr.clone())?
+			oracles
+				.add_named(name.clone())
+				.composite_mle(n_vars, inner_oracles, expr.clone())?
+		}
+		ColumnDef::RepeatingTransparent { poly } => {
+			let prev = oracles
```
jimpo-ulvt, line 354: This breaks the nice property that all columns in a table have consecutive oracle IDs in the oracle set. I think it'd be better to do an initial pass over all tables to add the transparents, then in a second pass add all of the table oracles.

jimpo-ulvt (changes requested):

crates/m3/src/builder/column.rs
```diff
@@ -138,4 +138,7 @@ pub enum ColumnDef<F: TowerField = B128> {
 		cols: Vec<ColumnIndex>,
 		expr: ArithExpr<F>,
 	},
+	RepeatingTransparent {
```
jimpo-ulvt, line 141: Rename to `Transparent`

jimpo-ulvt (approved):

PR #105, opened 2025-03-07
Expose computed columns from the M3 Table interface

### LINKED ISSUE #104: Expose computed columns from the M3 Table interface
We recently introduced the `binius_m3` crate, which is the new and preferred API for writing M3 constraint systems. This provides an abstraction layer over the polynomial oracles in the core library.

https://github.com/IrreducibleOSS/binius/pull/61 introduced a new multilinear oracle time for CompositeMLE polynomial oracles. We should expose access to this through the `binius_m3` crate. The naming in the `binius_m3` crate can be higher-level, describing the functionality to developers, whereas naming in the core library refers to how it's implemented. In that spirit, I propose calling these "computed columns" in the M3 crate.

Thus, the function signature to implement would be 

```rust
impl<'a, F: TowerField> TableBuilder<'a, F> {
	pub fn add_computed<FSub, const VALUES_PER_ROW: usize>(
		&mut self,
		name: impl ToString,
		expr: Expr<FSub, VALUES_PER_ROW>,
	) -> Col<FSub, VALUES_PER_ROW>
	where
		FSub: TowerField,
		F: ExtensionField<FSub>,
	{
		todo!()
	}
}
```

This is effectively the same method signature as `TableBuilder::add_linear_combination`, but in the implementation it will not require the expression to be linear.

TomWambsgans:
Attempt to fix #104

The test I wrote is quite ugly, I only did it to ensure I wasn’t missing something trivial. Unless contraindicated, I think it's better to remove it. (Otherwise I should improve it)

As for the PR itself, it seems pretty straightforward, I hope I don't miss anything. 2 remarks:
- I introduced 2 new functions for `ArithExpr`: `vars()` and `vars_sorted()`, and there may be confusion because `arithm_circuit.n_vars()` returns the max variable index + 1, which is different in general from `arithm_circuit.vars.len()` (`Var(0) * Var(5)` -> `n_vars()` = 6 but `vars().len()` = 2). Maybe we should rename ?
- While writing the test, I could not create an m3 `Expr` based on columns / field elements with different tower heights. Is there a raison not to allow this ?

### DIFF
--- a/crates/math/src/arith_expr.rs
+++ b/crates/math/src/arith_expr.rs
@@ -2,7 +2,6 @@
 
 use std::{
 	cmp::Ordering,
-	collections::HashSet,
 	fmt::{self, Display},
 	iter::{Product, Sum},
 	ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign},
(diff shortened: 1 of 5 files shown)


bergkvist:
> * While writing the test, I could not create an m3 `Expr` based on columns / field elements with different tower heights. Is there a raison not to allow this ?

Mixing tower heights in expressions should be allowed, but mixing `VALUES_PER_ROW` should not, so we should indeed improve the Expr/Column interface to allow this.

bergkvist (dismissed):

crates/math/src/arith_expr.rs
```diff
@@ -49,6 +49,21 @@ impl<F: Field> ArithExpr<F> {
 		}
 	}
 
+	/// The variable indices the expression contains, sorted and deduplicated.
+	pub fn used_vars(&self) -> Vec<usize> {
```
jimpo-ulvt, line 53: I introduced something very similar here https://github.com/IrreducibleOSS/binius/pull/110/files#diff-20087514335714fcc8576d3c7826cc49d447edf70a006417f5ac206cd60c96c8R274. Please use that method instead.

jimpo-ulvt (changes requested):

crates/math/src/arith_expr.rs
TomWambsgans: done

jimpo-ulvt (approved):

bergkvist (approved):

PR #109, opened 2025-03-09
Switch to fixed-size byte-sliced fields

GraDKh:
With the current implementation of byte-sliced fields, the packed fields for the different scalars have different sizes, so we can't implement `PackedExtension` for them. In this PR we replace them with 3D byte-sliced fields that always have an array of 16 8-bit packed AES fields with the size of a SIMD register.

UPD: After the conversation with @jimpo-ulvt it was decided to leave the smaller fields we used to have before


crates/hash/src/vision/permutation.rs
```diff
@@ -158,6 +158,8 @@ impl Vision32bPermutation {
 
 #[inline]
 fn add_packed_768<P: PackedField, const N: usize>(a: &mut [P; N], b: &[P; N]) {
+	println!("add_packed_768, b = {:?}", b);
```
jimpo-ulvt, line 161: stray print

jimpo-ulvt (approved):

crates/hash/src/vision/permutation.rs
GraDKh: Removed

jimpo-ulvt (approved):

PR #111, opened 2025-03-09
Make `Vision32Digest` implement `digest::Digest` trait and use Keccak padding scheme

GraDKh:
We want `Visiom32Digest` to implement `digest::Digest`, but this is impossible with the previous padding scheme (write the message length at the very beginning) since `digest::Digest` must be able to be created without parameters.
It was decided to switch to the Keccak padding scheme instead.

### DIFF
--- a/crates/hash/src/vision/digest.rs
+++ b/crates/hash/src/vision/digest.rs
@@ -1,21 +1,16 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::marker::PhantomData;
-
 use binius_field::{
-	as_packed_field::{PackScalar, PackedType},
-	make_aes_to_binary_packed_transformer, make_binary_to_aes_packed_transformer,
-	packed::set_packed_slice,
-	underlier::{Divisible, WithUnderlier},
-	AESTowerField32b, AesToBinaryTransformation, BinaryField, BinaryField32b, BinaryField8b,
-	BinaryToAesTransformation, ExtensionField, Field, PackedAESBinaryField8x32b,
-	PackedBinaryField8x32b, PackedExtension, PackedExtensionIndexable, PackedField,
-	PackedFieldIndexable,
+	linear_transformation::Transformation, make_aes_to_binary_packed_transformer,
+	make_binary_to_aes_packed_transformer, underlier::WithUnderlier, AesToBinaryTransformation,
+	BinaryField8b, BinaryToAesTransformation, PackedAESBinaryField8x32b, PackedBinaryField8x32b,
+	PackedExtensionIndexable, PackedField, PackedFieldIndexable,
 };
+use digest::{consts::U32, FixedOutput, HashMarker, OutputSizeUser, Update};
 use lazy_static::lazy_static;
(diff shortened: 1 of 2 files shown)


crates/hash/src/vision/digest.rs
```diff
@@ -38,306 +33,154 @@ lazy_static! {
...
-			HashError::NotEnoughData {
-				committed: 20,
-				hashed: 3,
-			}
-		);
-		assert!(successful_failure);
-	}
+/// Fill the data using Keccak padding scheme.
+#[inline(always)]
+fn fill_padding(data: &mut [u8]) {
```
Anex007, line 130: With the keccak padding scheme we start with a single 1 bit t followed zero or more 0 bits then a final 1 bit. Since we are filling rates with byte alignment the start should be `0x80` and the ending should be `0x01`

crates/hash/src/vision/digest.rs
```diff
@@ -38,306 +33,154 @@ lazy_static! {
...
-		let mut hasher: Vision32b<BinaryField32b> = Vision32b::new(1);
-		hasher.update([BinaryField32b::new(u32::from_le_bytes([
-			0xde, 0xad, 0xbe, 0xef,
-		]))]);
-		let out = hasher.finalize().unwrap();
+		let mut hasher = VisionHasherDigest::default();
+		let data = [0xde, 0xad, 0xbe, 0xef];
+		hasher.update(data);
+		let out = hasher.finalize();
 		// This hash is retrieved from a modified python implementation with the proposed padding and the changed mds matrix.
```
Anex007, line 151: NIT: outdated comment

crates/hash/src/vision/digest.rs
```diff
@@ -1,21 +1,16 @@
...
+	make_binary_to_aes_packed_transformer, underlier::WithUnderlier, AesToBinaryTransformation,
+	BinaryField8b, BinaryToAesTransformation, PackedAESBinaryField8x32b, PackedBinaryField8x32b,
+	PackedExtensionIndexable, PackedField, PackedFieldIndexable,
 };
+use digest::{consts::U32, FixedOutput, HashMarker, OutputSizeUser, Update};
 use lazy_static::lazy_static;
 
 use super::permutation::PERMUTATION;
-use crate::{permutation::Permutation, FixedLenHasher, HashError};
+use crate::permutation::Permutation;
```
Anex007, line 13: NIT: In the comment for `Vision32bPermutation` we only permutations implemented for `PackedAESBinaryField8x32b` and `ByteSlicedAES32x32b`

Anex007 (changes requested):
The padding bits seems to be reversed

crates/hash/src/vision/digest.rs
GraDKh: Yes, that's right changed both in this PR and the one for models too (it contained the same bug)

Anex007 (approved):
Looks good to me

jimpo-ulvt (dismissed):

PR #112, opened 2025-03-10
[m3] Use shared witness data for packed columns

bergkvist:
This reduces boilerplate when working with packed columns (which is typical when flushing columns after working with bit-representations). You now can decide whether to fill the packed or unpacked column. They both try to borrow the same RefCell mutably, so if you try to borrow both at the same time your code will panic.

### DIFF
--- a/crates/m3/src/builder/witness.rs
+++ b/crates/m3/src/builder/witness.rs
@@ -3,6 +3,7 @@
 use std::{
 	cell::{Ref, RefCell, RefMut},
 	iter, slice,
+	sync::Arc,
 };
 
 use anyhow::ensure;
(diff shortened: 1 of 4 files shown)


crates/m3/src/builder/witness.rs
```diff
@@ -386,8 +455,7 @@ impl<U: UnderlierType, F: TowerField> TableWitnessIndexSegment<'_, U, F> {
 		F: ExtensionField<FSub>,
 	{
 		let col = self
-			.cols
-			.get(col.table_index)
+			.get_col_data(col.table_index)
 			.ok_or_else(|| Error::MissingColumn(col.id()))?;
 		let col_ref = col.try_borrow().map_err(Error::WitnessBorrow)?;
```
jimpo-ulvt, line 460: Can you improve the error message for `WitnessBorrow` & `WitnessBorrowMut` to explain that packed columns are aliased?

crates/m3/src/builder/witness.rs
```diff
@@ -151,6 +117,52 @@ impl<'cs, 'alloc, U: UnderlierType, F: TowerField> WitnessIndex<'cs, 'alloc, U,
 	}
 }
 
+fn multilin_poly_from_underlier_data<U, F>(
+	data: &[U],
+	n_vars: usize,
+	tower_height: usize,
+) -> Arc<dyn MultilinearPoly<PackedType<U, F>> + Send + Sync + '_>
+where
+	F: TowerField
```
jimpo-ulvt, line 126: I think you can just make this not generic over `F` and instead just return `Arc<dyn MultilinearPoly<PackedType<U, B128>>`. The type bounds force that to be true already.

jimpo-ulvt (commented):
Nice job! It's kind of wacky to implement but a great memory optimization.

jimpo-ulvt (approved):

PR #113, opened 2025-03-11
[m3] Create gadget for the Rijndael S-box

jimpo-ulvt:
This is the first gadget towards building the Grøstl hash function with the M3 crate.

The main trickiness is constraining the hash over the canonical B8 tower field instead of the AES tower.

The input to the S-box is an `Expr` instead of a `Col` because the input can be a column with a round constant added, and it's more efficient to treat that as an expression than a linear combination column.


Anex007 (approved):
Nice

onesk (approved):

PR #119, opened 2025-03-12
Implement Multidigest for Vision32

GraDKh:
- Add `MultiDigest` trait for digests that calculate some fixed number of hash functions at a time
 - On top of it implement `ParallelDigest` trait. The naive solution would be to iterate over all chunk with the same index and calculate all hashes in parallel. However this is not cache friendly, better to calculate the fixed number of hashes that Multidigest implementation supports. That's why we need an `ParallelDigestSource` instead of Array2D or some kind of iterator.

### DIFF
--- a/crates/hash/src/multi_digest.rs
+++ b/crates/hash/src/multi_digest.rs
@@ -1,0 +1,117 @@
+// Copyright 2024-2025 Irreducible Inc.
+
+use std::{array, mem::MaybeUninit};
+
+use digest::Output;
+
+/// An object that efficiently computes `N` instances of a cryptographic hash function
+/// in parallel.
+pub trait MultiDigest<const N: usize>: Default + Clone {
+	/// The corresponding non-parallelized hash function.
+	type Digest: digest::Digest;
+
+	/// Returns the number of parallel instances that are computed.
+	fn parallel_instances() -> usize {
+		N
+	}
+
+	/// Create new hasher instance which has processed the provided data.
+	fn new_with_prefix(data: impl AsRef<[u8]>) -> Self {
+		let mut hasher = Self::default();
+		hasher.update([data.as_ref(); N]);
+		hasher
+	}
+
+	/// Process data, updating the internal state.
+	/// The number of rows in `data` must be equal to `parallel_instances()`.
+	fn update(&mut self, data: [&[u8]; N]);
+
+	/// Process input data in a chained manner.
+	#[must_use]
+	fn chain_update(self, data: [&[u8]; N]) -> Self {
+		let mut hasher = self;
+		hasher.update(data);
+		hasher
+	}
+
+	/// Write result into provided array and consume the hasher instance.
+	fn finalize_into(self, out: &mut [MaybeUninit<Output<Self::Digest>>; N]);
+
+	/// Write result into provided array and reset the hasher instance.
+	fn finalize_into_reset(&mut self, out: &mut [MaybeUninit<Output<Self::Digest>>; N]);
+
+	/// Reset hasher instance to its initial state.
+	fn reset(&mut self);
+
+	/// Compute hash of `data`.
+	fn digest(data: [&[u8]; N], out: &mut [MaybeUninit<Output<Self::Digest>>; N]);
+}
+
+/// Data source for the multidigest computation.
+///
+/// Provides an interface to the collection of data chunks by hash index and chunk index.
+/// All data chunks at the same index must be of the same length for all hashes.
+pub trait ParallelDigestSource {
+	/// Number of hashes to calculate
+	fn hashes(&self) -> usize;
+
+	/// Number of data chunks to calculate each hash digest
+	fn chunks(&self) -> usize;
+
+	/// Get the data chunk
+	fn get_chunk(&self, hash: usize, chunk: usize) -> &[u8];
+}
+
+pub trait ParallelDigest {
+	/// The corresponding non-parallelized hash function.
+	type Digest: digest::Digest;
+
+	/// Create new hasher instance which has processed the provided data.
+	fn new_with_prefix(data: impl AsRef<[u8]>) -> Self;
+
+	/// Calculate the digest of multiple hashes of data chunks of the same length
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>>;
+}
+
+struct ParallelDigestImpl<D: MultiDigest<N>, const N: usize>(D);
+
+impl<D: MultiDigest<N>, const N: usize> ParallelDigest for ParallelDigestImpl<D, N> {
+	type Digest = D::Digest;
+
+	fn new_with_prefix(data: impl AsRef<[u8]>) -> Self {
+		Self(D::new_with_prefix(data))
+	}
+
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>> {
+		let hashes = source.hashes();
+		let chunks = source.chunks();
+		let multihashes = hashes / N;
+		(0..multihashes)
+			.flat_map(move |i| {
+				let mut hasher = self.0.clone();
+				for chunk in 0..chunks {
+					let data = array::from_fn(|j| source.get_chunk(i * N + j, chunk));
+					hasher.update(data);
+				}
+
+				let mut out = array::from_fn(|_| MaybeUninit::uninit());
+				hasher.finalize_into(&mut out);
+				out.into_iter().map(|out| unsafe { out.assume_init() })
+			})
+			.chain((0..hashes % N).map(move |i| {
+				use digest::Digest as _;
+				let mut hasher = D::Digest::new();
+
+				for chunk in 0..chunks {
+					hasher.update(source.get_chunk(multihashes * N + i, chunk));
+				}
+				hasher.finalize()
+			}))
+	}
+}
--- a/crates/hash/src/vision/digest.rs
+++ b/crates/hash/src/vision/digest.rs
@@ -1,16 +1,20 @@
 // Copyright 2024-2025 Irreducible Inc.
 
+use std::{array, mem::MaybeUninit};
+
 use binius_field::{
 	linear_transformation::Transformation, make_aes_to_binary_packed_transformer,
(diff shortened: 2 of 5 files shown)


crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
...
+
+	fn new_with_prefix(data: impl AsRef<[u8]>) -> Self {
+		Self(D::new_with_prefix(data))
+	}
+
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>> {
+		let hashes = source.hashes();
```
djadjka, line 92: maybe better use `n_hashes`, `n_chunks`

djadjka (approved):
lgtm

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright 2024-2025 Irreducible Inc.
+
+use std::{array, mem::MaybeUninit};
+
+use digest::Output;
+
+/// An object that efficiently computes `N` instances of a cryptographic hash function
+/// in parallel.
```
jimpo-ulvt, line 8: Docstring should explain how this is designed for SIMD hash function implementation and how it relates to `ParallelDigest`. If this is mostly designed to be implemented but not called directly, docstring should say that.

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
...
+	/// Calculate the digest of multiple hashes of data chunks of the same length
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>>;
+}
+
+struct ParallelDigestImpl<D: MultiDigest<N>, const N: usize>(D);
+
+impl<D: MultiDigest<N>, const N: usize> ParallelDigest for ParallelDigestImpl<D, N> {
```
jimpo-ulvt, line 81: Please write a unit test for this in this module (independent of Vision).

crates/hash/src/vision/digest.rs
```diff
@@ -135,12 +139,194 @@ fn fill_padding(data: &mut [u8]) {
...
+	) {
+		let out = unsafe { slice_assume_init_mut(out) }
+			.try_into()
+			.expect("array is 32 elements");
+		self.finalize(out);
+		self.reset();
+	}
+
+	fn reset(&mut self) {
+		for v in &mut self.state {
```
jimpo-ulvt, line 303: nit: `self.state.fill(PackedField::zero())`

crates/hash/src/vision/digest.rs
```diff
@@ -135,12 +139,194 @@ fn fill_padding(data: &mut [u8]) {
...
+			.try_into()
+			.expect("array is 4 elements");
+		self.finalize(out);
+	}
+
+	fn finalize_into_reset(
+		&mut self,
+		out: &mut [MaybeUninit<digest::Output<Self::Digest>>; HASHES_PER_BYTE_SLICED_PERMUTATION],
+	) {
+		let out = unsafe { slice_assume_init_mut(out) }
```
jimpo-ulvt, line 295: This will have to get marked as a nightly feature, I guess?

crates/hash/src/vision/digest.rs
```diff
@@ -135,12 +139,194 @@ fn fill_padding(data: &mut [u8]) {
...
+				*state_element = TRANS_CANONICAL_TO_AES.transform(&canonical_data);
+			}
+		}
+
+		PERMUTATION.permute_mut(state);
+	}
+
+	fn finalize(
+		&mut self,
+		out: &mut [digest::Output<VisionHasherDigest>; HASHES_PER_BYTE_SLICED_PERMUTATION],
```
jimpo-ulvt, line 200: Why doesn't this one take `MaybeUninit`?

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
...
+	type Digest = D::Digest;
+
+	fn new_with_prefix(data: impl AsRef<[u8]>) -> Self {
+		Self(D::new_with_prefix(data))
+	}
+
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>> {
```
jimpo-ulvt, line 91: I was hoping that instead of returning an iterator, this would accept the output buffer of maybeuninits up from to reduce copies. I think we can do that with this method: https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3C%26mut+%5BT%5D%3E-for-%26mut+%5BT;+N%5D

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
...
+use digest::Output;
+
+/// An object that efficiently computes `N` instances of a cryptographic hash function
+/// in parallel.
+pub trait MultiDigest<const N: usize>: Default + Clone {
+	/// The corresponding non-parallelized hash function.
+	type Digest: digest::Digest;
+
+	/// Returns the number of parallel instances that are computed.
+	fn parallel_instances() -> usize {
```
jimpo-ulvt, line 14: This method seems unused and not useful (because the caller knows `N`)

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	fn digest(
+		&self,
+		source: &impl ParallelDigestSource,
+	) -> impl Iterator<Item = Output<Self::Digest>> {
+		let hashes = source.hashes();
+		let chunks = source.chunks();
+		let multihashes = hashes / N;
+		(0..multihashes)
```
jimpo-ulvt, line 95: This should be rayonized as well

jimpo-ulvt (changes requested):

crates/hash/src/vision/digest.rs
GraDKh: This is a function from `stackalloc`

crates/hash/src/multi_digest.rs
GraDKh: Applied

crates/hash/src/multi_digest.rs
GraDKh: Added

crates/hash/src/multi_digest.rs
GraDKh: Done

crates/hash/src/multi_digest.rs
GraDKh: Removed

crates/hash/src/multi_digest.rs
GraDKh: Just in case I've added a separate method that uses parallel iterations

crates/hash/src/vision/digest.rs
GraDKh: Used `bytemuck::fill_zeroes`

crates/hash/src/vision/digest.rs
GraDKh: Agree, modified

crates/hash/src/multi_digest.rs
GraDKh: Updated the `ParallelDigest` trait and `BinaryMerkeTree` now uses it instead of `digest::Digest` (`ParallelDigest` is implemented for `digest::Digest`, so we can use all hashers as well)

GraDKh:
Updated the PR, use `ParallelDigest` trait in the `BinaryMerkleTree`, so now we can compare the performance of Groestl256, single Vision32 and byte-sliced Vision32. With the byte-sliced implementation we have pretty much the same throughput as for Groestl

```
slow/merkle_tree/Grøstl-256/17 log elems size 16xBinaryField128b leaf
                        time:   [94.662 ms 95.078 ms 95.659 ms]
                        thrpt:  [334.52 MiB/s 336.57 MiB/s 338.04 MiB/s]

slow/merkle_tree/Vision-32/17 log elems size 16xBinaryField128b leaf
                        time:   [203.34 ms 204.13 ms 204.97 ms]
                        thrpt:  [156.12 MiB/s 156.76 MiB/s 157.37 MiB/s]

slow/merkle_tree/Vision-32-Parallel/17 log elems size 16xBinaryField128b leaf
                        time:   [97.930 ms 99.894 ms 102.11 ms]
                        thrpt:  [313.40 MiB/s 320.34 MiB/s 326.76 MiB/s]
```

jimpo-ulvt:
@GraDKh Could you split out the trait def & Merkle tree changes into a new MR that can be reviewed & merged first? Then all the Vision changes to be merged in a follow-up.

GraDKh:
Updated PR, now it contains Vision32 changes only

crates/hash/src/vision/digest.rs
```diff
@@ -162,12 +169,199 @@ fn fill_padding(data: &mut [u8]) {
...
+	) {
+		for row in &data {
+			debug_assert_eq!(row.len(), RATE_AS_U8);
+		}
+
+		for state_element_index in 0..2 {
+			let data_offset = state_element_index * HASHES_PER_BYTE_SLICED_PERMUTATION;
+
+			// Transpose the data to have the byte-sliced order
+			// TODO: Use transposition function here as soon as it is merged in.
```
jimpo-ulvt, line 204: Isn't this merged now?

crates/hash/src/vision/digest.rs
```diff
@@ -162,12 +169,199 @@ fn fill_padding(data: &mut [u8]) {
...
+			for row in 0..HASHES_PER_BYTE_SLICED_PERMUTATION {
+				fill_padding(&mut self.buffer[row][self.filled_bytes..]);
+			}
+
+			Self::permute(&mut self.state, array::from_fn(|i| &self.buffer[i]));
+		} else {
+			Self::permute(&mut self.state, array::from_fn(|_| &*PADDING_BLOCK));
+		}
+
+		// TODO: Use transposition function here as soon as it is merged in.
```
jimpo-ulvt, line 243: Is this ready?

crates/hash/src/vision/digest.rs
GraDKh: Yes, this code was written initially at the moment when transposition was not ready

jimpo (approved):

PR #122, opened 2025-03-14
[benches] Add micro benchmarks for main field operations

Anex007:
- This MR adds the main binary and unary operations for various packed fields that we care about as listed in #CRY-202

### DIFF
--- a/crates/field/benches/main_field_binary_ops.rs
+++ b/crates/field/benches/main_field_binary_ops.rs
@@ -1,0 +1,71 @@
+// Copyright 2025 Irreducible Inc.
+
+mod packed_field_utils;
+
+use std::ops::Mul;
+
+use binius_field::{
+	arch::OptimalUnderlier, as_packed_field::PackedType, AESTowerField128b, AESTowerField16b,
+	AESTowerField32b, AESTowerField64b, AESTowerField8b, BinaryField128b, BinaryField128bPolyval,
+	BinaryField16b, BinaryField32b, BinaryField64b, BinaryField8b, ByteSlicedAES32x128b,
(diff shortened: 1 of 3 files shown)


GraDKh (approved):
Generally speaking I'm OK with these changes. Possibly we could add `OptimalXXX` packed fields in the list in `packed_field_utils.rs` and feature-gate it, so that we could use the benchmarks we already have for that, just pass a special feature flag to bench optimal packed fields only. But I'm not insisting on that.

crates/field/benches/main_field_binary_ops.rs
```diff
@@ -0,0 +1,71 @@
...
+
+trait SelfMul: Mul<Self, Output = Self> + Sized {}
+
+impl<T: Mul<Self, Output = Self> + Sized> SelfMul for T {}
+
+fn mul_main<T: SelfMul>(lhs: T, rhs: T) -> T {
+	lhs * rhs
+}
+
+benchmark_packed_operation!(
```
jimpo-ulvt, line 38: This is missing 1b fields

crates/field/benches/main_field_unary_ops.rs
```diff
@@ -0,0 +1,70 @@
...
+
+fn invert<T: PackedField>(val: T) -> T {
+	val.invert_or_zero()
+}
+
+fn square<T: PackedField>(val: T) -> T {
+	val.square()
+}
+
+benchmark_packed_operation!(
```
jimpo-ulvt, line 36: This is missing 1b fields

jimpo-ulvt (approved):

PR #127, opened 2025-03-17
Fix`channel_id_bound`

4l0n50:
The field `channel_id_bound` was not being updated when adding a new channel to m3's constraint system.

### DIFF
--- a/crates/m3/src/builder/constraint_system.rs
+++ b/crates/m3/src/builder/constraint_system.rs
@@ -142,6 +142,7 @@ impl<F: TowerField> ConstraintSystem<F> {
 
 	pub fn add_channel(&mut self, name: impl ToString) -> ChannelId {
 		let id = self.channels.len();
+		self.channel_id_bound += 1;
 		self.channels.push(Channel {
 			name: name.to_string(),
 		});


jimpo-ulvt:
This field shouldn't be on the struct at all, we should just use `channels.len()`. Looks like maybe a leftover from when the `ConstraintSystemBuilder` was removed.

bergkvist:
> My original problem is that this assertion in `channel::validate_witness` fails when there's more than one channel:
> 
> ```rust
> 		if channel_id > max_channel_id {
> 			return Err(Error::ChannelIdOutOfRange {
> 				max: max_channel_id,
> 				got: channel_id,
> 			});
> 		}.
> ```
> 
> I think the issue is that the `validate_witness` functions only know about the flushes. Or at least, I haven't found any reference to the channels for computing `channels.len()` there.

This is where you want to change to `channels.len().saturating_sub(1)`:
https://github.com/IrreducibleOSS/binius/blob/b21e9c8872f5ac9ce7193363682ecdfe42cd07d7/crates/m3/src/builder/constraint_system.rs#L281-L287

This should allow you to remove the channel_id_bound from the m3 ConstraintSystem struct here:
https://github.com/IrreducibleOSS/binius/blob/b21e9c8872f5ac9ce7193363682ecdfe42cd07d7/crates/m3/src/builder/constraint_system.rs#L33-L38

4l0n50:
I see, thanks @bergkvist !

jimpo-ulvt (approved):

bergkvist (approved):

PR #134, opened 2025-03-18
[hash] Import Grøstl implementation from RustCrypto

jimpo-ulvt:
This fully implements the `Groestl256` Digest in the `binius_hash` crate. The code is written to have an architecture-abstracted trait called `GroestlShortInternal` that the Digest is built on. The portable implementation of it copies code from https://github.com/RustCrypto/hashes with minimal modifications and tests directly against it with a proptest.

In a subsequent PR, I will refactor the AVX512-optimized Grøstl code to expose this interface so that it gets used.


crates/maybe_rayon/Cargo.toml
```diff
@@ -14,5 +14,5 @@ itertools.workspace = true
 rayon = { optional = true, workspace = true }
 
 [features]
-default = ["rayon"]
+default = []# "rayon"]
```
Anex007, line 17: Shouldn't rayon be enabled by default

Anex007 (commented):
two small comments

crates/maybe_rayon/Cargo.toml
jimpo-ulvt: Oops, good catch

Anex007 (dismissed):
LGTM

onesk (approved):

onesk (approved):

PR #137, opened 2025-03-19
Modify hasher challenger

josephjohnston:
I've made two complementary changes to `HasherChallenger`. 

- When switching from sampling to observing, the number of samples that were performed is no longer recorded. Recording the number of samples was proposed by @djadjka such that sampling different numbers of challenges in one round implies different challenges in subsequent rounds. Due to the second change in this PR described next, recording the number of samples is no longer necessary. It also much-simplifies writing the challenger in zCray-lang.
- When initializing the `HasherChallenger`, previously we initialized with the empty hasher, and do so in a `default` method. Now we instead pass in an `initial_digest` to initialize the hasher, and do so in a `new` method. This initial digest will be the digest of some form of hashing both the serialized constraint system, and the statement being proven. For now, without this hashing in place yet, all instantiations of the `HasherChallenger` continue to call `default` rather than `new`, and within `default` I call `new` passing in the digest of the empty string.
Making this change entails initializing the `HasherChallenger` in sampling state rather than observing state.

Here's why the second change safely allows for the first change.
If the verifier performs a different number of samples in one round, that means the constraint system must be different, since there is one deterministic verifier algorithm for each constraint system. If the constraint system is different, that means the `initial_hash` initializing the `HasherChallenger` is different. If the `HasherChallenger` is initialized differently, the challenges sampled from it will be different. Therefore, we have achieved the same effect as @djadjka, enforcing that sampling two different numbers of challenges in one rounds implies different challenges in subsequent rounds.

### DIFF
--- a/crates/core/src/fiat_shamir/hasher_challenger.rs
+++ b/crates/core/src/fiat_shamir/hasher_challenger.rs
@@ -13,6 +13,7 @@ use super::Challenger;
 /// Challenger type which implements `[Buf]` that has similar functionality as `[CanSample]`
 #[derive(Debug, Default)]
 pub struct HasherSampler<H: Digest> {
+	sample_count: usize,
 	index: usize,
 	buffer: Output<H>,
 	hasher: H,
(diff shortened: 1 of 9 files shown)


jimpo-ulvt (changes requested):

djadjka (commented):
If I understand correctly, it helps to handle cases with different constraint systems, but it is not allowed to detect protocols bugs that occur when sample is called a different number of times in prove and verify.

josephjohnston:
re: recording the number of samples performed before switching back to sampling: @djadjka I see, good point, I didn't think about that. I can see this is very useful for debugging, but @benediamond points out we shouldn't change the protocol for debugging purposes. Do you feel detecting these bugs is useful beyond debugging? We can add it back if so. any thoughts @jimpo?

djadjka:
I can offer to wrap it in an additional debug feature. It would allow us to catch errors when needed, but not apply it to the protocol in production.

josephjohnston:
> I can offer to wrap it in an additional debug feature. It would allow us to catch errors when needed, but not apply it to the protocol in production.

that would be great, thank you. (or describe how and I can try)

jimpo:
I like the current behavior that the transcript state is bound uniquely to the sequence of observe and sample calls and is not malleable (in the sense that different numbers of sample calls can produce the same transcript state). @benediamond I'm inclined to close this and keep the current behavior.

jimpo:
@josephjohnston The transcript should really take in the full specification of the interaction sequence and then the transcript interactions would check against that for debugging purposes.

crates/core/src/fiat_shamir/hasher_challenger.rs
```diff
@@ -21,75 +22,141 @@ pub struct HasherSampler<H: Digest> {
...
+		let mut hasher = H::new();
+		Digest::update(&mut hasher, &initial_digest);
+
+		Self {
+			interactions: Vec::new(),
+			observer_or_sampler: ObserverOrSampler::Sampler(HasherSampler {
+				sample_count: 0,
+				hasher,
+				index: 0,
+				buffer: initial_digest,
```
djadjka, line 83: I'm not sure, but do we really need to use initial_digest as a buffer after adding it to the hasher?

djadjka (approved):
LGTM

jimpo:
@josephjohnston Please remove the runtime checks and just use trace logs to help debug, as discussed.

PR #144, opened 2025-03-20
Add ParallelDigest trait and use it in BinaryMerkleTree

GraDKh:


crates/core/src/constraint_system/prove.rs
```diff
@@ -74,7 +74,8 @@ where
 	U: ProverTowerUnderlier<Tower>,
 	Tower: ProverTowerFamily,
 	Tower::B128: PackedTop<Tower>,
-	Hash: Digest + BlockSizeUser + FixedOutputReset,
+	DomainFactory: EvaluationDomainFactory<FDomain<Tower>>,
```
jimpo-ulvt, line 77: This seems to be a bad merge

crates/core/src/merkle_tree/binary_merkle_tree.rs
```diff
@@ -30,11 +34,11 @@ pub fn build<F, H, C>(
...
 	elements: &[F],
 	batch_size: usize,
-) -> Result<BinaryMerkleTree<Output<H>>, Error>
+) -> Result<BinaryMerkleTree<Output<H::Digest>>, Error>
 where
 	F: TowerField,
-	H: Digest + BlockSizeUser + FixedOutputReset,
-	C: PseudoCompressionFunction<Output<H>, 2> + Sync,
+	H: ParallelDigest<Digest: BlockSizeUser + FixedOutputReset>,
+	C: PseudoCompressionFunction<Output<H::Digest>, 2> + Sync,
```
jimpo-ulvt, line 41: This would be a separate change, but we should also have a trait for parallel compression.

crates/core/src/merkle_tree/binary_merkle_tree.rs
```diff
@@ -177,38 +181,66 @@ where
...
 	F: TowerField,
-	H: Digest + BlockSizeUser + FixedOutputReset,
+	H: ParallelDigest<Digest: BlockSizeUser + FixedOutputReset>,
 	ParIter: IndexedParallelIterator<Item: IntoIterator<Item = F>>,
 {
-	digests
-		.par_iter_mut()
-		.zip(iterated_chunks)
-		.for_each_init(H::new, |hasher, (digest, elems)| {
+	digests.par_iter_mut().zip(iterated_chunks).for_each_init(
```
jimpo-ulvt, line 241: Do we need rayon here? I figured this is the point of the ParallelDigest trait

jimpo-ulvt (commented):
Very nice job!!

crates/core/src/constraint_system/prove.rs
GraDKh: Fixed

crates/core/src/merkle_tree/binary_merkle_tree.rs
GraDKh: I've updated `ParallelDigest` trait one more time to specialize it for our case when we have a single chunk of hash data that can be serialized to some buffer (which is actually an iterator over field elements in the `BinaryMerkleTree` case). Now we can use parallel digest in both places there

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,315 @@
+// Copyright 2024-2025 Irreducible Inc.
```
jimpo-ulvt, line 1:
```suggestion
// Copyright 2025 Irreducible Inc.
```

jimpo-ulvt: nit: ^ new code can begin copyright this year

crates/hash/src/multi_digest.rs
```diff
@@ -0,0 +1,315 @@
...
+	/// Write result into provided array and consume the hasher instance.
+	fn finalize_into(self, out: &mut [MaybeUninit<Output<Self::Digest>>; N]);
+
+	/// Write result into provided array and reset the hasher instance.
+	fn finalize_into_reset(&mut self, out: &mut [MaybeUninit<Output<Self::Digest>>; N]);
+
+	/// Reset hasher instance to its initial state.
+	fn reset(&mut self);
+
+	/// Compute hash of `data`.
```
jimpo-ulvt, line 55: Does this require all `N` slices to be the same length?

crates/hash/src/multi_digest.rs
GraDKh: Fixed

crates/hash/src/multi_digest.rs
GraDKh: Yes, added a note about it to the method doc

PR #145, opened 2025-03-20
feat: enable flushes with custom selectors on `m3`

arthurpaulino:
Add `pull_selected` and `push_selected` functions to `TableBuilder` and `TablePartition`, allowing API consumers to provide selectors for flushes.

The `StepDown` selector is used as a fallback in case one is not provided.

### DIFF
--- a/crates/m3/src/builder/table.rs
+++ b/crates/m3/src/builder/table.rs
@@ -264,6 +264,34 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
 		self.table.partition_mut(1).push(channel, cols);
 	}
 
+	pub fn pull_selected<FSub>(
+		&mut self,
+		channel: ChannelId,
+		cols: impl IntoIterator<Item = Col<FSub>>,
+		selector: Col<FSub>,
+	) where
+		FSub: TowerField,
+		F: ExtensionField<FSub>,
+	{
+		self.table
+			.partition_mut(1)
+			.pull_selected(channel, cols, selector);
+	}
+
+	pub fn push_selected<FSub>(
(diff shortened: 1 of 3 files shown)


crates/m3/src/builder/table.rs
```diff
@@ -241,15 +241,43 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
...
 	pub fn push<FSub>(&mut self, channel: ChannelId, cols: impl IntoIterator<Item = Col<FSub>>)
 	where
 		FSub: TowerField,
 		F: ExtensionField<FSub>,
 	{
-		self.table.partition_mut(1).push(channel, cols);
+		self.table.partition_mut(1).push(channel, cols, None);
+	}
+
+	pub fn pull_custom<FSub>(
```
jimpo-ulvt, line 255: Please rename to `pull_selected` and `push_selected`

crates/m3/src/builder/table.rs
```diff
@@ -317,27 +345,46 @@ impl<F: TowerField> TablePartition<F> {
 		});
 	}
 
-	pub fn pull<FSub>(&mut self, channel: ChannelId, cols: impl IntoIterator<Item = Col<FSub>>)
-	where
+	pub fn pull<FSub>(
```
jimpo-ulvt, line 348: Please leave the fn signature as it is, without a selector. Same for `push`.

The change to `flush` is OK because it's private.

jimpo-ulvt (changes requested):

jimpo-ulvt (approved):

PR #149, opened 2025-03-21
[benches] Extract gpa prodcheck for nightly

Anex007:
- This MR adds the prodcheck necessary for nightly benchmarks

### DIFF
--- a/.github/workflows/nightly-benchmark.yml
+++ b/.github/workflows/nightly-benchmark.yml
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron: "0 3 * * *"
   workflow_dispatch:  # Allows manual trigger
+  push:
+    branches:
+      - anexj/gpa_micro_benchmarks
 
 permissions:
   contents: write
@@ -13,7 +16,6 @@ permissions:
 jobs:
   benchmark:
     name: Benchmark for ${{ matrix.os }}
-    if: github.ref_name == 'main'
     container: rustlang/rust:nightly
     runs-on: ${{ matrix.os }}
     continue-on-error: true
--- a/.github/workflows/nightly-benchmark.yml
+++ b/.github/workflows/nightly-benchmark.yml
@@ -45,8 +45,17 @@ jobs:
           --bench main_field_unary_ops \
           --bench sumcheck \
           --bench large_transform \
-          --bench nightly_prodcheck \
           --bench binary_merkle_tree'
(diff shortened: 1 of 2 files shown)


crates/core/benches/nightly_prodcheck.rs
```diff
@@ -0,0 +1,44 @@
...
+		N_VARS,
+		N_CLAIMS,
+		"gpa_binary_128b",
+		EvaluationOrder::LowToHigh,
+		c,
+	);
+}
+
+fn bench_byte_sliced_aes_256b(c: &mut Criterion) {
+	// TODO: this benchmarks should account for the byte sliced transposition time
```
jimpo-ulvt, line 33: Nah, doesn't need to

crates/core/benches/nightly_prodcheck.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2025 Irreducible Inc.
```
jimpo-ulvt, line 1: This doesn't make sense that it's a different file. Can't you just filter for the benchmarks or bench group from the CI runner?

jimpo-ulvt (changes requested):

crates/core/benches/nightly_prodcheck.rs
Anex007: Yeah you can run it with filter, two issues I found:
1) In CI, you need to have a second command that just runs with just a filter (This spams the bencher reports, since we have 6 reports per run 3 testbeds single and multi threaded)
2) This is how the filter looks like if we were to keep the original file: `cargo bench --bench prodcheck -- "gpa_(polyval_128b|binary_128b|byte_sliced_aes_256b)/.*=20"`.

milosbackonja (commented):
Should this be a separate bencher command or part of existing?
Would separate step make more sense here?

Anex007:
> Should this be a separate bencher command or part of existing? Would separate step make more sense here?

This is because of the way criterion commands work, If you apply the filter to all benchmarks it would only run just the gpa benchmarks and exclude everything else

jimpo-ulvt (dismissed):

Anex007:
Tested the github action here: https://github.com/IrreducibleOSS/binius/actions/runs/14032932910
This MR is ready

milosbackonja (approved):

jimpo (approved):

PR #150, opened 2025-03-21
Fix propagation of no default features (to restore the support of building without `rayon`)

GraDKh:


crates/m3/Cargo.toml
```diff
@@ -9,8 +9,8 @@ anyhow.workspace = true
 binius_core = { path = "../core" }
 binius_field = { path = "../field" }
 binius_math = { path = "../math" }
-binius_maybe_rayon = { path = "../maybe_rayon" }
-binius_utils = { path = "../utils" }
+binius_maybe_rayon = { path = "../maybe_rayon", default-features = false }
```
jimpo, line 12: Shouldn't this crate now have a `rayon` or `parallel` feature? How do we build it with rayon?

jimpo: With this MR, it looks like the `m3` crate always gets built without rayon and there's no way to enable it. Is that not true?

jimpo: Ah, nevermind https://doc.rust-lang.org/cargo/reference/features.html#feature-unification.

jimpo (approved):

PR #151, opened 2025-03-21
Save duplication partial evaluation in evalcheck and ring_switch

djadjka:
Both evalcheck and ring_switch may perform the same evaluate_partial_high operation on multilinear polynomial witnesses. In evalcheck, this occurs when computing projections. The multilinears then get stored in a memoization map to make sure they are not recomputed in evalcheck. Then in ring_switch, the prover computes the partial high evaluations to compute the tensor algebra elements sent. The prover save the computation of the projection, if available, so it is not recomputed during ring switching.

### DIFF
--- a/crates/core/src/protocols/evalcheck/prove.rs
+++ b/crates/core/src/protocols/evalcheck/prove.rs
@@ -17,7 +17,8 @@ use super::{
 	evalcheck::{EvalcheckMultilinearClaim, EvalcheckProof},
 	subclaims::{
 		add_composite_sumcheck_to_constraints, calculate_projected_mles, composite_sumcheck_meta,
-		fill_eq_witness_for_composites, MemoizedQueries, ProjectedBivariateMeta,
+		fill_eq_witness_for_composites, memoize_partial_evals, MemoizedQueries,
+		ProjectedBivariateMeta,
 	},
 	EvalPoint, EvalPointOracleIdMap,
 };
@@ -30,7 +31,7 @@ use crate::{
 		packed_sumcheck_meta, process_packed_sumcheck, process_shifted_sumcheck,
 		shifted_sumcheck_meta,
 	},
-	witness::MultilinearExtensionIndex,
+	witness::{MultilinearExtensionIndex, MultilinearWitness},
 };
 
 /// A mutable prover state.
@@ -61,7 +62,8 @@ where
 	projected_bivariate_claims: Vec<EvalcheckMultilinearClaim<F>>,
 
 	new_sumchecks_constraints: Vec<ConstraintSetBuilder<F>>,
-	memoized_queries: MemoizedQueries<PackedType<U, F>, Backend>,
+	pub memoized_queries: MemoizedQueries<PackedType<U, F>, Backend>,
+	pub memoized_partial_evals: EvalPointOracleIdMap<MultilinearWitness<'b, PackedType<U, F>>, F>,
 	backend: &'a Backend,
 }
 
@@ -91,6 +93,7 @@ where
 			claims_without_evals_dedup: EvalPointOracleIdMap::new(),
 			projected_bivariate_claims: Vec::new(),
 			memoized_queries: MemoizedQueries::new(),
+			memoized_partial_evals: EvalPointOracleIdMap::new(),
 			backend,
 			incomplete_proof_claims: EvalPointOracleIdMap::new(),
 		}
@@ -238,11 +241,19 @@ where
 
 		for (claim, meta, projected) in izip!(
 			std::mem::take(&mut self.projected_bivariate_claims),
-			projected_bivariate_metas,
+			projected_bivariate_metas.clone(),
 			projected_mles
 		) {
 			self.process_sumcheck(claim, meta, projected)?;
 		}
+
+		memoize_partial_evals(
+			&projected_bivariate_metas,
+			&self.projected_bivariate_claims,
+			self.oracles,
+			self.witness_index,
+			&mut self.memoized_partial_evals,
+		);
 
 		// Step 4: Find and return the proofs of the original claims.
 
@@ -611,7 +622,7 @@ where
 				eval,
 				self.witness_index,
 				&mut self.new_sumchecks_constraints,
-				projected.expect("projected is required by shifted oracle"),
+				projected,
 			),
 
 			MultilinearPolyVariant::Packed(packed) => process_packed_sumcheck(
@@ -622,7 +633,7 @@ where
 				eval,
 				self.witness_index,
 				&mut self.new_sumchecks_constraints,
-				projected.expect("projected is required by packed oracle"),
+				projected,
 			),
 
 			MultilinearPolyVariant::Composite(composite) => {
--- a/crates/core/src/protocols/evalcheck/subclaims.rs
+++ b/crates/core/src/protocols/evalcheck/subclaims.rs
@@ -24,12 +24,12 @@ use binius_math::{
 use binius_maybe_rayon::prelude::*;
 use binius_utils::bail;
 
-use super::{error::Error, evalcheck::EvalcheckMultilinearClaim};
+use super::{error::Error, evalcheck::EvalcheckMultilinearClaim, EvalPointOracleIdMap};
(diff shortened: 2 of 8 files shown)


jimpo-ulvt:
@onesk Can you review please?

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -61,7 +62,8 @@ where
 	projected_bivariate_claims: Vec<EvalcheckMultilinearClaim<F>>,
 
 	new_sumchecks_constraints: Vec<ConstraintSetBuilder<F>>,
-	memoized_queries: MemoizedQueries<PackedType<U, F>, Backend>,
+	pub memoized_queries: MemoizedQueries<PackedType<U, F>, Backend>,
```
onesk, line 65: nit: I personally would prefer to add a method that would consume the prover and return memoized queries/partial evals (ideally as a tuple or custom structure to package them together, given that we treat both atomically and never split them apart, just pass on to the ring switching code). Making things `pub` technically allows mutable access which is inappropriate here.

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -238,12 +241,20 @@ where
 
 		for (claim, meta, projected) in izip!(
 			std::mem::take(&mut self.projected_bivariate_claims),
-			projected_bivariate_metas,
+			projected_bivariate_metas.clone(),
```
onesk, line 244: nit: `meta` is `Copy` so you could've used `&projected_bivariate_metas` there instead of cloning (accompanied by `&meta` pattern match in `for`).

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -328,29 +331,54 @@ where
...
+		.for_each(|(claim, meta)| {
+			let inner_id = meta.inner_id.unwrap();
+			if matches!(oracles.oracle(inner_id).variant, MultilinearPolyVariant::Committed)
+				&& meta.projected_id.is_some()
+			{
+				let eval_point = claim.eval_point[meta.projected_n_vars..].to_vec().into();
+
+				let projected_id = meta.projected_id.expect("checked above");
+
+				let projected = witness_index.get_multilin_poly(projected_id).unwrap();
```
onesk, line 375: no `unwrap` please

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -137,7 +137,7 @@ pub fn composite_sumcheck_meta<F: TowerField>(
 		multiplier_id: oracles.add_transparent(EqIndPartialEval::new(eval_point.to_vec()))?,
 		inner_id: None,
 		projected_id: None,
-		projected_n_vars: eval_point.len(),
+		projected_n_vars: 0,
```
onesk, line 140: either leave a comment that this field is not referenced in composite case, or introduce a separate meta type without this field

crates/core/src/ring_switch/prove.rs
```diff
@@ -61,7 +64,13 @@ where
 	let mixing_coeffs = MultilinearQuery::expand(&mixing_challenges).into_expansion();
 
 	// For each evaluation point prefix, send one batched partial evaluation.
-	let tensor_elems = compute_partial_evals::<_, _, _, Tower, _>(system, witnesses, backend)?;
+	let tensor_elems = compute_partial_evals::<_, _, _, Tower, _>(
+		system,
+		witnesses,
+		memoized_queries,
```
onesk, line 70: Commented elsewhere - imo it's preferable to coalesce `memoized_queries` & `memoized_partial_evals` into one `memo` struct

onesk (changes requested):

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -611,7 +620,7 @@ where
 				eval,
 				self.witness_index,
 				&mut self.new_sumchecks_constraints,
-				projected.expect("projected is required by shifted oracle"),
```
jimpo, line 614: Why remove this and pass the option?

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -222,28 +223,36 @@ where
...
 		for (claim, meta, projected) in izip!(
 			std::mem::take(&mut self.projected_bivariate_claims),
-			projected_bivariate_metas,
+			&projected_bivariate_metas,
 			projected_mles
 		) {
 			self.process_sumcheck(claim, meta, projected)?;
 		}
 
+		memoize_partial_evals(
```
jimpo, line 248: Why is this called here? The results don't get used in this method AFAICT?

crates/core/src/protocols/evalcheck/subclaims.rs
```diff
@@ -328,33 +332,28 @@ where
...
+		.map(|(claim, meta)| match (meta.inner_id, meta.projected_id) {
+			(Some(inner_id), Some(_)) => {
+				let inner_multilin = witness_index.get_multilin_poly(inner_id)?;
+				let eval_point = &claim.eval_point[meta.projected_n_vars..];
+				let query = memoized_queries
+					.full_query_readonly(eval_point)
+					.ok_or(Error::MissingQuery)?;
+				Ok(Some(
+					backend
+						.evaluate_partial_high(&inner_multilin, query.to_ref())
```
jimpo, line 344: This partial evaluation doesn't seem to be added to the memoized map? Am I missing something?

crates/core/src/ring_switch/prove.rs
```diff
@@ -136,18 +142,30 @@ where
...
-				let partial_eval =
-					witnesses[*committed_idx].evaluate_partial_high(suffix_query.to_ref())?;
-				TowerTensorAlgebra::new(
-					suffix.kappa,
+				let suffix_desc = &system.suffix_descs[*suffix_desc_idx];
+
+				let elems = if let Some(partial_eval) =
+					memoized_data.partial_eval(eval_claim.id, Arc::as_ref(&suffix_desc.suffix))
+				{
+					PackedField::iter_slice(
```
jimpo, line 152: Deduplicate this between branches. Consider an interface for MemoizedData that [tells instead of asking](https://martinfowler.com/bliki/TellDontAsk.html) (like `MemoizedData::full_query`).

jimpo (changes requested):

crates/core/src/protocols/evalcheck/subclaims.rs
jimpo: ^ This is major. Reusing this computation is the main purpose of the issue.

crates/core/src/ring_switch/prove.rs
djadjka: Not sure that I can apply it here, in the first case we look up partial_eval by ID, and in the second case it's a evaluate_partial_high MultilinearPoly.
Yes, we could do something like `partial_eval(witness: M, witness_id: OracleId, eval_point: &[F])`, but that seems a bit overloaded.

crates/core/src/protocols/evalcheck/prove.rs
djadjka: This method builds a map from (id, eval_point) to MultilinearWitness, which will be used in ring_switch instead of storing evals separately

jimpo: Then this code should be in `ring_switch` if it's only used in `ring_switch`

djadjka: Now it's part of MemoizedData, so I think it's the right place

crates/core/src/protocols/evalcheck/subclaims.rs
djadjka: yes, but at the end of evalcheck we build map `(id, eval_point) -> MultilinearWitness`, and for ring_switch it's just result of `evaluate_partial_high`, just to avoid duplication of evals

jimpo (approved):

onesk (approved):

PR #161, opened 2025-03-25
Introduce a struct for a batch of rows of the same length

GraDKh:
It appeared that we are creating that big batch of values when proving Vision32 that just checking if all the rows in the batch are of the same length takes signifficant amount of time since we are doing it for each `batch_evaluate` call. in this PR we introduce a special struct that guarantees that all the rows are of the same length by checking it at the moment of creation. Also I like this approach more because this struct helps self-describe the argument of `batch_evaluate` .
For proving Vision32 this saves about 15% of overall time.

### DIFF
--- a/crates/math/src/rows_batch.rs
+++ b/crates/math/src/rows_batch.rs
@@ -31,11 +31,10 @@ impl<'a, T> RowsBatch<'a, T> {
 	}
 
 	#[inline(always)]
-	pub fn get_ref(&self) -> RowsBatchRef<'a, '_, T> {
+	pub fn get_ref(&self) -> RowsBatchRef<'_, T> {
 		RowsBatchRef {
 			rows: self.rows.as_slice(),
 			row_len: self.row_len,
-			_pd: std::marker::PhantomData,
 		}
 	}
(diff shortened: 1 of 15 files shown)


jimpo:
@onesk Please review

crates/math/src/rows_batch.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2025 Irreducible Inc.
+
+/// This struct represents a batch of rows, each row having the same length equal to `row_len`.
+pub struct RowsBatch<'a, T> {
```
onesk, line 4: Is this struct intended to be used in tests only? It has an allocation inside, and as fair as I can see it's not used in any of the hotspots?

onesk (dismissed):
A very nice optimization!

jimpo:
Nice idea!

I managed to simplify the lifetime signature of RowsBatchRef [here](https://github.com/IrreducibleOSS/binius/commit/11b896a497af001de7fdd834cfe25355599e6268).

jimpo (approved):

PR #169, opened 2025-03-31
[MultilinearExtensionIndex] Use PackedField instead of PackedType<U, F>

djadjka:
Initially, Underlier was supposed to be an internal trait, but over time it has spread into various parts of the project where it could be avoided. So I plan to open a series of PRs to clean the project up.


GraDKh (dismissed):

jimpo (approved):
Beautiful, I love this!

jimpo (approved):

PR #171, opened 2025-03-31
[piop] Make piop compiler work with high-to-low sumcheck

jimpo:
This is a quick modification to make the FRI-Binius PCS use high-to-low sumcheck. The trick is simply that we bit-reverse the multilinear coefficients before committing. This has the effect of bit-reversing their encodings as well, and everything works pretty nicely.

I'm not that happy about re-introducing the dependency on p3-util or `PackedFieldIndexable`, but this was the fastest way to get it working. @GraDKh maybe you have ideas how to fix that?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,6 +107,7 @@ lazy_static = "1.5.0"
 paste = "1.0.15"
 proc-macro2 = "1.0.81"
 proptest = "1.2.0"
+p3-util = { version = "0.1.0", git = "https://github.com/Plonky3/Plonky3" }
 quote = "1.0.36"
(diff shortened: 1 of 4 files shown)


GraDKh (approved):
I think it's ok to merge it, but add a TODO and task to remove the reference to the plonky3. That utility function seems to be pretty simple, we can have our own implementation that works on any type of collection that allows swapping scalar elements, including a slice of packed fields

onesk (approved):
I'm okay with merging it, as long as `PackedFieldIndexable` gets dropped eventually - otherwise its a deal breaker.

jimpo:
Created issue for cleanup https://github.com/IrreducibleOSS/binius/issues/172

PR #174, opened 2025-04-01
[field] Friendlier function for getting field bases

jimpo:
The `ext_basis<FExt, FSub>` function is now an easy way to get field extension bases.

### DIFF
--- a/crates/m3/src/gadgets/hash/groestl.rs
+++ b/crates/m3/src/gadgets/hash/groestl.rs
@@ -20,9 +20,7 @@ use binius_field::{
 };
 use bytemuck::Pod;
 
-use crate::builder::{
-	upcast_col, upcast_expr, Col, Expr, TableBuilder, TableWitnessSegment, B1, B8,
-};
+use crate::builder::{upcast_col, Col, Expr, TableBuilder, TableWitnessSegment, B1, B8};
 
 /// The first row of the circulant matrix defining the MixBytes step in Grøstl.
 const MIX_BYTES_VEC: [u8; 8] = [0x02, 0x02, 0x03, 0x04, 0x05, 0x03, 0x05, 0x07];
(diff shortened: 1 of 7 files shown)


crates/m3/src/gadgets/hash/groestl.rs
```diff
@@ -326,9 +327,7 @@ struct SBox<const V: usize> {
 
 impl<const V: usize> SBox<V> {
 	pub fn new(table: &mut TableBuilder, input: Expr<B8, V>) -> Self {
-		let b8_basis: [_; 8] = array::from_fn(|i| {
-			<B8 as ExtensionField<B1>>::basis(i).expect("i in range 0..8; extension degree is 8")
-		});
+		let b8_basis: [_; 8] = array::from_fn(ext_basis::<B8, B1>);
```
4l0n50, line 330: If this is used also used in other places we could declare something like
```rust
pub(crate) const B8_B1_BASIS: std::sync::LazyLock<[B8; 8]> = std::sync::LazyLock::new(|| {
   array::from_fn(ext_basis::<B8, B1>);
});
```

crates/m3/src/gadgets/hash/groestl.rs
```diff
@@ -326,9 +327,7 @@ struct SBox<const V: usize> {
...
 		let pack_b8 = move |bits: [Expr<B1, V>; 8]| {
```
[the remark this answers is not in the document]
jimpo: Hmm, I'm fine moving it from a closure to a method, but I don't think it should be a `From` impl. A function like `pack_b8` should work fine and is a bit more descriptive.

crates/m3/src/gadgets/hash/groestl.rs
jimpo: This call should be very fast and is not in a very hot loop, so I'm in favor of leaving it rather than using a lazy static.

sai-deng (approved):
Thanks!

4l0n50 (approved):

PR #176, opened 2025-04-01
Flush with const

rsen-irreducible:
#175 Support for flushing into channel without requiring OracleIds.


crates/circuits/src/builder/constraint_system.rs
```diff
@@ -110,13 +110,26 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
...
 		U: PackScalar<BinaryField1b>,
 	{
-		let n_vars = self.log_rows(oracle_ids.clone())?;
+		//We assume there is atleast one non constant in the collection of oracle ids.
+		let non_const_oracles: Vec<usize> = oracle_ids
+			.clone()
+			.into_iter()
+			.filter_map(|id| match id {
+				OracleOrConst::Oracle(oracle_id) => Some(oracle_id),
+				OracleOrConst::Const {
```
jimpo, line 125: nit: `OracleOrConst::Const { .. }`

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -110,13 +110,26 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
...
 		count: usize,
-		oracle_ids: impl IntoIterator<Item = OracleId> + Clone,
+		oracle_ids: impl IntoIterator<Item = OracleOrConst<F>> + Clone,
 		multiplicity: u64,
 	) -> anyhow::Result<()>
 	where
 		U: PackScalar<BinaryField1b>,
 	{
-		let n_vars = self.log_rows(oracle_ids.clone())?;
+		//We assume there is atleast one non constant in the collection of oracle ids.
```
jimpo, line 119:
```suggestion
		// We assume there is at least one non constant in the collection of oracle ids.
```

crates/core/src/constraint_system/verify.rs
```diff
@@ -466,9 +467,22 @@ pub fn make_flush_oracles<F: TowerField>(
...
+						flush.oracles.iter().copied().filter_map(|id| match id {
+							OracleOrConst::Oracle(oracle_id) => Some(oracle_id),
+							OracleOrConst::Const {
+								base: _,
+								tower_level: _,
+							} => None,
+						});
+
+					let first_oracle = non_const_oracles
+						.clone()
```
djadjka, line 480: Do we really need a clone here? I might be wrong, but it looks like we can just call next() and continue comparing without using skip(1).

jimpo (commented):
Nice! This is somewhat cumbersome to call from the circuit API, but I think it's still a useful feature/optimization (compared to creating constant transparents).

crates/core/src/constraint_system/verify.rs
```diff
@@ -488,16 +502,37 @@ pub fn make_flush_oracles<F: TowerField>(
...
+						.copied()
+						.zip(mixing_powers.iter())
+						.filter_map(|(id, coeff)| match id {
+							OracleOrConst::Const {
+								base,
+								tower_level: _,
+							} => Some(base * coeff),
+							OracleOrConst::Oracle(_oracle_id) => None,
+						})
+						.sum();
```
djadjka, line 517: In our code style, we prefer to use turbofish syntax (e.g., sum::\<F\>())

djadjka (commented):
Could you please add some tests?
Also, as I remember, the ticket featured ExpBase, so it would be great to update that too. 🙂

crates/circuits/src/builder/constraint_system.rs
rsen-irreducible: Done

crates/circuits/src/builder/constraint_system.rs
jimpo: If you change it to `OracleOrConst::Const { .. }`. cargo fmt leaves it alone. I just tried.

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -143,18 +156,31 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
...
 	) -> anyhow::Result<()> {
-		let oracles = oracle_ids.into_iter().collect::<Vec<_>>();
-		let log_rows = self.log_rows(oracles.iter().copied())?;
+		//We assume there is atleast one non constant in the collection of oracle ids.
+		let non_const_oracles: Vec<usize> = oracle_ids
+			.clone()
+			.into_iter()
+			.filter_map(|id| match id {
+				OracleOrConst::Oracle(oracle_id) => Some(oracle_id),
+				OracleOrConst::Const {
```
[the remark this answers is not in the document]
rsen-irreducible: I'll follow the latter convention.

crates/circuits/src/builder/constraint_system.rs
djadjka: Looks like you lost this =)

crates/core/src/constraint_system/verify.rs
rsen-irreducible: Your suggestion is much cleaner. The iterator would just need to be declared mutable.

crates/core/src/constraint_system/verify.rs
```diff
@@ -466,9 +467,16 @@ pub fn make_flush_oracles<F: TowerField>(
...
+					let mut non_const_oracles =
+						flush.oracles.iter().copied().filter_map(|id| match id {
+							OracleOrConst::Oracle(oracle_id) => Some(oracle_id),
+							_ => None,
+						});
+
+					let first_oracle = non_const_oracles.next().ok_or(Error::EmptyFlushOracles)?;
+					let n_vars = oracles.n_vars(first_oracle);
+
+					for oracle_id in non_const_oracles.clone().skip(1) {
```
djadjka, line 479: You call `next()`, so non_const_oracles no longer contains the first element, and you don't need clone here as I see 
Also, you can try the same approach as [here](https://github.com/IrreducibleOSS/binius/pull/176/files#diff-d8801fee06d94b89752c8c9511ef03f055db4899b72444173b2c8f3ee28fc2d4R147)

rsen-irreducible: Oh right my bad.

crates/circuits/src/builder/constraint_system.rs
```diff
@@ -452,3 +472,91 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
 		Ok(log_rows)
 	}
 }
+
+#[cfg(test)]
```
djadjka, line 476: Pls move tests to the crates/circuits/src/lib.rs

djadjka (commented):
Looks good, please update ExoBase too

crates/circuits/src/builder/constraint_system.rs
rsen-irreducible: ExpBase has already been updated, moved the test case.

djadjka (approved):
LGTM

jimpo (approved):

PR #178, opened 2025-04-02
[evalcheck] Deduplicate Evalcheck claims

Anex007:
This is still work in progress but currently greedy evalcheck verification passes without issues, but ring swtich verification fails.


jimpo:
This needs a lot more documentation explaining the algorithm.

PR #187, opened 2025-04-03
New transparents added for lookups.

rsen-irreducible:
Additional transparents for:

1. Mapping a point in the boolean hypercube to the appropriate B32 element.
2. Mapping a query (i || j) in the boolean hypercube to (i || j || ( i AND j )).
3. Mapping a query (i || j || cin) in the boolean hypercube to (i || j || (i+j bits sans overflow when added as integers) || cin || cout).


jimpo:
Closing as we'll soon implement with a new lookup pattern in the m3 crate

PR #188, opened 2025-04-04
Support for no switchover in sumchecks

onesk:
* Instead of always taking an `M: MultilinearPoly` transparent, `ProverState` ingests `SumcheckMultilinear` directly
* `EqIndSumcheckProverBuilder` was augmented with `no_switchover` constructor, which allows specifying folded `Vec<P>` large field representations by value

Rationale:
* GPA sumchecks operate on large field _unique_ witnesses; using `immediate_switchover_heuristic` still allocates 1.5x witness data, which is totally avoidable when witness is passed by value
* (continuation of the above) Shorter-than-power-of-two folded multilinears are explicitly supported by the HAL, which is useful for optimizing GPAs coming from plain lookups
* In zerocheck univariate round we currently skip at most 7 rounds, which results in folded multilinear still being larger than witness column for 1-bit oracles; this change unlocks potential optimization of passing the Lagrange interpolation inner product query to the follow up `EqIndSumcheck` for pre-switchover usage.

### DIFF
--- a/crates/core/src/constraint_system/prove.rs
+++ b/crates/core/src/constraint_system/prove.rs
@@ -771,16 +771,21 @@ where
 			nonzero_scalars_prefixes.push(entry.nonzero_scalars_prefix);
 		}
 
-		let prover = EqIndSumcheckProverBuilder::new(backend)
-			.with_nonzero_scalars_prefixes(&nonzero_scalars_prefixes)
-			.build(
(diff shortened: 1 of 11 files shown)


jimpo:
@djadjka Please review

djadjka (approved):

jimpo (approved):

PR #189, opened 2025-04-04
Arbitrary multilinear const suffix in sumcheck

onesk:
> Depends on #188 

Attempts to implement more efficient GPA sumchecks for plain lookups proved that existing zero extend multilinear support in sumcheck is inadequate - things like "offset bivariate products" of the form `(a+offset)*(b+offset)-offset` pollute too much verifier code for what is essentially a prover optimization.

Hence this PR replaces `nonzero_scalars_suffix` params with `const_suffix`, which is a pair `(F, usize)`. Sumcheck folding/round calculation and `EqIndSumcheck` const eval suffix determination are augmented to support this functionality.


djadjka (dismissed):
LGTM

jimpo (approved):

PR #192, opened 2025-04-04
Truncated GPA prover

### LINKED ISSUE #54: [gkr_gpa] Avoid copying the input layer data
See [this TODO comment](https://github.com/IrreducibleOSS/binius/blob/c035caef0bda2d5e187c362cb79ea776f0f1cc21/crates/core/src/protocols/gkr_gpa/gkr_gpa.rs#L43).

It should be possible to avoid copying the input layer data. This may require adding the multilinear poly to the Witness struct.

onesk:
> Depends on #189

GPA prover has been refactored to accept (potentially) truncated large field witnesses by value instead of references to `MultilinearPoly` impls. This fixes #54 as well as avoids unneeded allocations during proving, but most importantly it unlocks const eval suffix optimizations in `EqIndSumcheck`, which is an important plain lookup optimization.

Unfortunately I had to remove the `first_round_eval_at_1s` optimization - while it allows to reuse one multiplication from witness construction during sumchecks, keeping it requires holding on to a copy of current layer until the challenges get sampled in the corresponding sumcheck and carefully implementing evaluation at multiplicative identity to not be slower than just doing sumcheck, which seems like a lot of effort for a perf improvement hard capped at 15%.


crates/core/src/protocols/gkr_gpa/gkr_gpa.rs
```diff
@@ -31,119 +25,85 @@ impl<F: Field> GrandProductClaim<F> {
...
-		} else {
-			let layer_size = 1 << (i - 1);
-
-			let first_half = PackedFieldStorage::new_inline(layer[0].iter().take(layer_size))?;
-			let second_half =
-				PackedFieldStorage::new_inline(layer[0].iter().skip(layer_size).take(layer_size))?;
-
-			Ok((first_half, second_half))
-		}
+	/// TODO: comment
```
djadjka, line 104: "TODO"

djadjka (approved):
LGTM. 
The only question I have is about cases where we can't split the truncated input layer into packed fields without a remainder — but I'm not sure it's relevant here.

jimpo (approved):
Nice simplification!

PR #193, opened 2025-04-04
Add underliers and arch-optimal aliases for byte-sliced fields

GraDKh:
Add underliers (without bit operations) for the byte sliced packed fields. To my opinion this gives many pros:
1. Get rid of macros to manually define the `PackedExtension` chain.
2. Allows implementing aliases to get the architecture-optimal byte-sliced fields in a pretty easy way.
3. As constraint system needs underlier defined it allows us getting initial end-to-end proving faster no matter if we decide to leave or remove the underlier there. I would prefer doing it in parallel with performance investigation instead of being blocked by that.

Also I switched to acr-optimal byte sliced fields in the benchmarks where we compare it against arch-optimal 'ordinary' packed fields.

### DIFF
--- a/crates/core/benches/prodcheck.rs
+++ b/crates/core/benches/prodcheck.rs
@@ -8,12 +8,12 @@ use binius_core::{
 	transcript::ProverTranscript,
 };
 use binius_field::{
-	arch::OptimalUnderlier,
+	arch::{OptimalUnderlier, OptimalUnderlierByteSliced},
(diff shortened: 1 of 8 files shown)


jimpo:
Longer term, this is something I will probably want to roll back. In the meantime, we still use Underliers in a bunch of places and I trust your judgement that this makes things easier.

jimpo (approved):

PR #194, opened 2025-04-04
[m3] Ability to define power-of-two-sized tables

jimpo:
This adds the option to define tables in M3 that are required to have power-of-two size. These tables will have selectors omitted. This also modifies the constraint proving protocol to skip the selected sumcheck when the table is power-of-two-sized.

**Why not just omit selectors if the table size in the instance happens to be a power of two?** This is a reasonable alternative. However, I ultimately want the property that the oracle IDs and types in the compiled constraint system are independent of the table sizes in the instance. This is necessary for reduction planning. Also, it seems unlikely that large tables that are not statically known to be power-of-two sized will end up being power-of-two sized.

### DIFF
--- a/crates/m3/src/builder/table.rs
+++ b/crates/m3/src/builder/table.rs
@@ -43,6 +43,10 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
 		}
 	}
 
+	pub fn require_power_of_two_size(&mut self) {
+		self.table.po2_sized = true;
+	}
+
 	pub fn with_namespace(&mut self, namespace: impl ToString) -> TableBuilder<'_, F> {
 		TableBuilder {
 			namespace: Some(self.namespaced_name(namespace)),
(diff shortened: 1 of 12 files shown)


crates/m3/src/builder/table.rs
```diff
@@ -320,6 +324,8 @@ pub struct Table<F: TowerField = B128> {
 	pub id: TableId,
 	pub name: String,
 	pub columns: Vec<ColumnInfo<F>>,
+	/// Whether the table size is required to be a power of two.
+	pub po2_sized: bool,
```
bergkvist, line 328: Maybe `requires_power_of_two_size` would be a more descriptive name here

bergkvist (approved):

crates/m3/src/builder/table.rs
jimpo: I changed to `power_of_two_sized`

PR #206, opened 2025-04-10
[fix] remove recursive BufMut and Buf coercion

alxkzmn:
## Issue
There was a trait resolution overflow caused by recursive `&mut` coercions when `BufMut` and `Buf` were taken as `impl Trait` and then re-borrowed in macro-derived `serialize()` and `deserialize()` methods.

## Steps To Reproduce
Try to compile the following unit test:
```
#[cfg(test)]
mod tests {
	use generic_array::typenum::U32;
	use rand::{rngs::StdRng, RngCore, SeedableRng};

	use super::*;

	#[test]
	fn test_generic_array_serialize_deserialize() {
		let mut rng = StdRng::seed_from_u64(0);

		let mut data = GenericArray::<u8, U32>::default();
		rng.fill_bytes(&mut data);

		let mut buf = Vec::new();
		data.serialize(&mut buf, SerializationMode::Native).unwrap();

		let data_deserialized =
			GenericArray::<u8, U32>::deserialize(&mut buf.as_slice(), SerializationMode::Native)
				.unwrap();
		assert_eq!(data_deserialized, data);
	}
}
```
The compilation would fail with a message similar to the following:
```
error[E0275]: overflow evaluating the requirement &mut BytesMut: BufMut
  |
  = help: consider increasing the recursion limit by adding a #![recursion_limit = "256"] attribute to your crate (...)
  = note: required for &mut &mut BytesMut to implement BufMut
  = note: 128 redundant requirements hidden
  = note: required for &mut &mut &mut &mut &mut &mut &mut &mut &mut &mut &mut ... to implement BufMut
```

## Fix
The derive macro and trait impls were updated to avoid moves and use a single borrow.
The corresponding implementations were updated accordingly.
Added a unit test for `ConstraintSystem` serialization/deserialization where the issue originally manifested.

### DIFF
--- a/crates/circuits/src/builder/constraint_system.rs
+++ b/crates/circuits/src/builder/constraint_system.rs
@@ -485,3 +485,34 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
 		Ok(log_rows)
 	}
 }
+
+#[cfg(test)]
+mod tests {
+	use binius_core::constraint_system::ConstraintSystem;
+	use binius_utils::{DeserializeBytes, SerializationMode, SerializeBytes};
+
+	use super::ConstraintSystemBuilder;
+	use crate::u32fib::u32fib;
+
+	#[test]
+	fn test_serialize_roundtrip() {
+		let mut builder = ConstraintSystemBuilder::new();
+		let log_size_1b = 14;
+		let _ = u32fib(&mut builder, "u32fib", log_size_1b).unwrap();
+
+		let prover_constraint_system = builder.build().unwrap();
+		let mut write_buf = vec![];
+		prover_constraint_system
+			.serialize(&mut write_buf, SerializationMode::Native)
+			.unwrap();
+		let constraint_system =
+			ConstraintSystem::deserialize(&mut write_buf.as_slice(), SerializationMode::Native)
+				.unwrap();
+		assert_eq!(constraint_system.exponents.len(), 0);
+		assert_eq!(constraint_system.flushes.len(), 0);
+		assert_eq!(constraint_system.max_channel_id, 0);
+		assert_eq!(constraint_system.non_zero_oracle_ids.len(), 0);
+		assert_eq!(constraint_system.table_constraints.len(), 2);
+		assert_eq!(constraint_system.oracles.size(), 9);
+	}
+}
--- a/crates/core/src/constraint_system/mod.rs
+++ b/crates/core/src/constraint_system/mod.rs
@@ -37,19 +37,19 @@ pub struct ConstraintSystem<F: TowerField> {
 
 impl DeserializeBytes for ConstraintSystem<BinaryField128b> {
 	fn deserialize(
-		mut read_buf: impl bytes::Buf,
+		read_buf: &mut dyn bytes::Buf,
 		mode: SerializationMode,
 	) -> Result<Self, SerializationError>
 	where
 		Self: Sized,
 	{
 		Ok(Self {
-			oracles: DeserializeBytes::deserialize(&mut read_buf, mode)?,
-			table_constraints: DeserializeBytes::deserialize(&mut read_buf, mode)?,
-			non_zero_oracle_ids: DeserializeBytes::deserialize(&mut read_buf, mode)?,
-			flushes: DeserializeBytes::deserialize(&mut read_buf, mode)?,
-			exponents: DeserializeBytes::deserialize(&mut read_buf, mode)?,
-			max_channel_id: DeserializeBytes::deserialize(&mut read_buf, mode)?,
+			oracles: DeserializeBytes::deserialize(read_buf, mode)?,
+			table_constraints: DeserializeBytes::deserialize(read_buf, mode)?,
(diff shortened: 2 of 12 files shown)


Anex007:
The issue comes from trying to unroll blanket implementations for `BufMut` and `Buf` behind mutable references, as this hits the recursion limit. Instead of using trait objects the simplest thing to do is to change the serialization to take a reference which implements `BufMut` and `Buf`. So the new `SerializeBytes` would look like
```
#[auto_impl(Box, &)]
pub trait SerializeBytes {
	fn serialize(
		&self,
		write_buf: &mut impl BufMut,
		mode: SerializationMode,
	) -> Result<(), SerializationError>;
}
```
In the `erased_serialize_bytes` you may have to wrap it inside a `Box`.  You will have to take a similar approach with `DeserializeBytes` to take in a reference of the type.

jimpo:
I can't reproduce this error. This test exists verbatim already in the main branch: https://github.com/IrreducibleOSS/binius/blob/15bd8f1e89a07755d6ca79971e36816a85ebbbfb/crates/utils/src/serialization.rs#L425

What platform are you on, and what Rust compiler version?

jimpo:
I dug some more and I believe this has to do with the implementation of `SerializeBytes` & `DeserializeBytes` on `ArithExpr`, which is a recursive datatype. Something about the recursive serialization on a recursive datatype triggers [this rustc issue](https://github.com/rust-lang/rust/issues/112991).

Interestingly, I found a workaround that lets us keep the method signature on the trait as is (which I prefer). When serializing a recursive datatype, it's OK if the buffer reference is cast to a trait object reference. Something like:

```rust
impl<F: Field> SerializeBytes for ArithExpr<F> {
	fn serialize<B: BufMut>(
		&self,
		mut write_buf: B,
		mode: SerializationMode,
	) -> Result<(), SerializationError> {
		match self {
			Self::Const(elem) => elem.serialize(write_buf, mode),
			Self::Var(i) => i.serialize(write_buf, mode),
			Self::Add(left, right) => {
				write_buf.put_u8(0x00);
				left.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
				Ok(())
			}
			Self::Mul(left, right) => {
				write_buf.put_u8(0x01);
				left.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
				Ok(())
			}
			Self::Pow(right, x) => {
				write_buf.put_u8(0x02);
				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
				Ok(())
			}
		}
	}
}
```

I think the ideal solution would be to do something like this inside the derive macro if possible. @bergkvist

jimpo (changes requested):
> I dug some more and I believe this has to do with the implementation of `SerializeBytes` & `DeserializeBytes` on `ArithExpr`, which is a recursive datatype. Something about the recursive serialization on a recursive datatype triggers [this rustc issue](https://github.com/rust-lang/rust/issues/112991).
> 
> Interestingly, I found a workaround that lets us keep the method signature on the trait as is (which I prefer). When serializing a recursive datatype, it's OK if the buffer reference is cast to a trait object reference. Something like:
> 
> ```rust
> impl<F: Field> SerializeBytes for ArithExpr<F> {
> 	fn serialize<B: BufMut>(
> 		&self,
> 		mut write_buf: B,
> 		mode: SerializationMode,
> 	) -> Result<(), SerializationError> {
> 		match self {
> 			Self::Const(elem) => elem.serialize(write_buf, mode),
> 			Self::Var(i) => i.serialize(write_buf, mode),
> 			Self::Add(left, right) => {
> 				write_buf.put_u8(0x00);
> 				left.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
> 				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
> 				Ok(())
> 			}
> 			Self::Mul(left, right) => {
> 				write_buf.put_u8(0x01);
> 				left.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
> 				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
> 				Ok(())
> 			}
> 			Self::Pow(right, x) => {
> 				write_buf.put_u8(0x02);
> 				right.serialize(&mut write_buf as &mut dyn BufMut, mode)?;
> 				Ok(())
> 			}
> 		}
> 	}
> }
> ```
> 
> I think the ideal solution would be to do something like this inside the derive macro if possible. @bergkvist

I thought about it more, and this actually seems like explainable compiler behavior, not https://github.com/rust-lang/rust/issues/112991. I recommend @Anex007 's solution, which is to accept `&mut impl BufMut` in the `serialize` and `deserialize` methods instead of `&mut dyn BufMut`, as you have implemented @alxkzmn.

pepyakin:
I will go ahead and close this PR because it looks stale. It's rather mechanical to clean up our review queue. Should you still want to land this feel free to reopen!

PR #207, opened 2025-04-10
[compute] Initial memory interface

jimpo:
@alsrdn and I learned in the course of pair programming that having `FSlice` traits that can subslice themselves is problematic. (It's complicated to explain: the subslices have a different type than the parent because they are parameterized by another lifetime (a subscope), and they're no way to additionally indicate to the type system the relationship at an abstract level).

Instead I made a `ComputeMemory` trait for manipulating memory references & slices. The eventual `ComputeLayer` trait will either have a `ComputeMemory` as a parent trait or associated type.

This also includes a dead-simple bump allocator implementation that is generic over `ComputeMemory`.

### DIFF
--- a/crates/compute/Cargo.toml
+++ b/crates/compute/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "binius_compute"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+
+[lints]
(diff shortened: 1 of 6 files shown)


alsrdn (approved):
LGTM

PR #210, opened 2025-04-11
[Evalcheck] Deduplication using advices

djadjka:
This PR introduces deduplication for EvalCheck claim prove/verify steps, which positively impacts both proof size and verification time.

The core idea is to leverage Advice — supplementary information the prover provides to the verifier. If a claim has already been proven/verified, the prover includes its position within the `round_claims` list. This allows the verifier to efficiently confirm the validity of a duplicate claim without the need for an additional search through the list.

The proving process has been updated and now consists of three key steps:
1. **Evaluation Memoization with modified BFS** 
   Since the order of processing stamps is not important, we take full advantage of multithreading  
   for improved performance.
2. **Prove multilinears**  
   Constructs both the proofs and the associated advice  
   in the same order that the verifier will process them.  
   This ensures consistency and correct indexing between prover and verifier.
3. **Processing `projected_bivariate_claims`**

### DIFF
--- a/crates/core/src/protocols/evalcheck/evalcheck.rs
+++ b/crates/core/src/protocols/evalcheck/evalcheck.rs
@@ -149,6 +149,10 @@ impl EvalcheckProofAdviceNumerics {
 	}
 }
 
+/// Additional information provided by the prover to the verifier.
+/// If the claim has already been proven/verified, the prover provides the position in the
+/// `round_claims` list. This allows the verifier to confirm the validity of a duplicate claim
+/// without performing an additional search through the list.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum EvalcheckProofAdvice {
 	HandleClaim,
--- a/crates/core/src/protocols/evalcheck/evalcheck.rs
+++ b/crates/core/src/protocols/evalcheck/evalcheck.rs
@@ -1,6 +1,7 @@
 // Copyright 2023-2025 Irreducible Inc.
 
 use std::{
+	hash::Hash,
 	ops::{Deref, Range},
 	slice,
 	sync::Arc,
@@ -15,7 +16,7 @@ use crate::{
 	transcript::{TranscriptReader, TranscriptWriter},
 };
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EvalcheckMultilinearClaim<F: Field> {
 	/// Virtual Polynomial Oracle for which the evaluation is claimed
 	pub id: OracleId,
@@ -308,7 +309,7 @@ impl<T: Clone, F: Field> Default for EvalPointOracleIdMap<T, F> {
 	}
 }
 
-#[derive(Debug, Clone, Hash, Eq)]
+#[derive(Debug, Clone, Eq)]
 pub struct EvalPoint<F: Field> {
 	data: Arc<[F]>,
 	range: Range<usize>,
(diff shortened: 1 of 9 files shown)


josephjohnston:
two comments
- I ended up writing it (for zcray) without `EvalcheckProofAdvice`, and instead adding an extra variant to `EvalcheckProof` called `DuplicateClaim(index)`. I just found it simpler that way.
- using `EvalcheckProofAdvice` the prover is able to influence the sequence of reductions and thus the sequence of observations and challenges. therefore we've stepped outside the realm of deterministic verifiers (I think we'd be the first proof system to do this). for now this just means we'll need to modify the challenger to once again observe the number of samples performed when switching from sampling to observing (our zcray challenger is currently written without doing this, because it's unnecessary for deterministic verifiers)

Anex007 (changes requested):
Looks Solid, just some comments.

crates/core/src/protocols/evalcheck/evalcheck.rs
```diff
@@ -85,17 +125,83 @@ impl EvalcheckNumerics {
...
-	evalcheck: &EvalcheckProof<F>,
+	evalcheck: &Option<EvalcheckProof<F>>,
 ) {
-	match evalcheck {
+	if evalcheck.is_none() {
+		transcript.write_bytes(&[EvalcheckNumerics::Duplicated as u8]);
+		return;
+	}
+
+	match evalcheck.clone().unwrap() {
```
Anex007, line 204: It's possible to avoid extra cloning here. Also `unwrap()`

crates/core/src/protocols/evalcheck/evalcheck.rs
```diff
@@ -110,64 +216,97 @@ pub fn serialize_evalcheck_proof<B: BufMut, F: TowerField>(
...
 	let mut ty = 0;
 	transcript.read_bytes(slice::from_mut(&mut ty))?;
 	let as_enum = EvalcheckNumerics::from(ty)?;
 
-	match as_enum {
-		EvalcheckNumerics::Transparent => Ok(EvalcheckProof::Transparent),
-		EvalcheckNumerics::Committed => Ok(EvalcheckProof::Committed),
-		EvalcheckNumerics::Shifted => Ok(EvalcheckProof::Shifted),
-		EvalcheckNumerics::Packed => Ok(EvalcheckProof::Packed),
+	if as_enum == EvalcheckNumerics::Duplicated {
```
Anex007, line 262: Maybe it's better to use serialization with `Option` as opposed to using `EvalcheckNumerics`. Might be worth it to do in a different PR

crates/core/src/protocols/evalcheck/tests.rs
```diff
@@ -539,38 +612,56 @@ fn test_evalcheck_zero_padded() {
 
 	let backend = make_portable_backend();
 	let mut prover_state = EvalcheckProver::new(&mut oracles, &mut witness_index, &backend);
-	let proof = prover_state.prove(vec![claim.clone()]).unwrap();
+	let ProofsWithAdvices { proofs, advices } = prover_state.prove(vec![claim.clone()]).unwrap();
```
Anex007, line 615: It worth it to assert the advice enums here to make sure deduplication is happening in the right order

crates/core/src/protocols/evalcheck/evalcheck.rs
```diff
@@ -216,8 +359,8 @@ impl<T: Clone, F: Field> Default for EvalPointOracleIdMap<T, F> {
 	}
 }
 
-#[derive(Debug, Clone)]
-pub struct EvalPoint<F> {
+#[derive(Debug, Clone, Eq)]
+pub struct EvalPoint<F: Field> {
```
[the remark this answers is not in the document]
djadjka: I think HashMap is complicated for the 1-2 elements =)

josephjohnston (approved):
verifier looks good to me

Anex007 (approved):

PR #211, opened 2025-04-12
[m3] Refactor away duplication of table size specification

jimpo:
Previously all table sizes needed to be specified when the witness index was created. There's enough information in the `fill_table_sequential` and `fill_table_parallel` calls to initialize the table indexes properly, so this change allows the table initialization to use the data from those calls. When creating the `Statement` for constraint system compilation, it's recommended to get the table information from the witness, or in the verifier's case, received from the prover.

### DIFF
--- a/crates/m3/src/builder/constraint_system.rs
+++ b/crates/m3/src/builder/constraint_system.rs
@@ -155,24 +155,8 @@ impl<F: TowerField> ConstraintSystem<F> {
 	pub fn build_witness<'cs, 'alloc, P: PackedField<Scalar = F>>(
 		&'cs self,
 		allocator: &'alloc Bump,
-		_statement: &Statement,
 	) -> Result<WitnessIndex<'cs, 'alloc, P>, Error> {
 		Ok(WitnessIndex::new(allocator, &self.tables))
-		// Ok(WitnessIndex {
-		// 	tables: self
-		// 		.tables
-		// 		.iter()
-		// 		.zip(&statement.table_sizes)
-		// 		.map(|(table, &table_size)| {
-		// 			let witness = if table_size > 0 {
-		// 				Some(TableWitnessIndex::new(allocator, table, table_size))
-		// 			} else {
-		// 				None
-		// 			};
-		// 			witness.transpose()
-		// 		})
-		// 		.collect::<Result<_, _>>()?,
-		// })
 	}
 
 	/// Compiles a [`CompiledConstraintSystem`] for a particular statement.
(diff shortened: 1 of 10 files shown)


crates/m3/src/builder/constraint_system.rs
```diff
@@ -155,23 +155,8 @@ impl<F: TowerField> ConstraintSystem<F> {
 	pub fn build_witness<'cs, 'alloc, P: PackedField<Scalar = F>>(
```
bergkvist, line 155: At this point I would consider removing this method, and just calling `WitnessIndex::new(&cs.tables, &allocator)` where `cs.build_witness(&allocator)` is being used now. If we want to keep build_witness around, its return type no longer needs to be a Result, and this section from the docstring should be updated/removed, since the statement is no longer passed as an argument:
> 	/// The statement includes information about the tables sizes, which this requires in order to
>	/// allocate the column data correctly. The created witness index needs to be populated before
>	/// proving.

crates/m3/src/builder/witness.rs
```diff
@@ -37,51 +39,120 @@ use crate::builder::multi_iter::MultiIterator;
...
+#[derive(Debug)]
 pub struct WitnessIndex<'cs, 'alloc, P = PackedType<OptimalUnderlier, B128>>
 where
 	P: PackedField,
 	P::Scalar: TowerField,
 {
-	pub tables: Vec<Option<TableWitnessIndex<'cs, 'alloc, P>>>,
+	allocator: &'alloc Bump,
+	/// Each entry is Left if the index hasn't been initialized & filled, and Right if it has.
+	tables: Vec<Either<&'cs Table<P::Scalar>, TableWitnessIndex<'cs, 'alloc, P>>>,
```
bergkvist, line 50: Another thing I would consider here, is to simply have the WitnessIndex hold an immutable reference to the ConstraintSystem, and have this be a `Vec<Option<TableWitnessIndex<'cs, 'alloc, P>>`

bergkvist (approved):
Nice usability improvement!

crates/m3/src/builder/constraint_system.rs
jimpo: Yeah, that makes sense. I'll leave this method to make it easier for existing code to migrate (there are some external codebases), but mark it deprecated. Will remove the Result type though, that doesn't make sense anymore.

crates/m3/src/builder/witness.rs
jimpo: I initially had something like that, but prefer having a single vector instead of two vectors that must be the same length by invariant.

bergkvist (approved):

Issue #88, opened 2025-03-05
Add DeserializeBytesB128 derive macro


bergkvist:
This should do the same as the `DeserializeBytes` macro, except it should substitute in `binius_field::BinaryField128b` for generic Field/TowerField parameters.

Examples of where this would be useful:
https://github.com/IrreducibleOSS/binius/blob/aac8231188833589af27e141c075154cb1a85316/crates/core/src/oracle/multilinear.rs#L290-L299

https://github.com/IrreducibleOSS/binius/blob/aac8231188833589af27e141c075154cb1a85316/crates/core/src/oracle/multilinear.rs#L462-L478

https://github.com/IrreducibleOSS/binius/blob/aac8231188833589af27e141c075154cb1a85316/crates/core/src/oracle/multilinear.rs#L492-L520

The reason this is desirable is that anything containing a `dyn MultivariatePoly<F>` will only implement deserialization for `F = BinaryField128b`

Right now, whenever an enum with manual deserialization is changed, the deserialization also needs to be updated correctly, which can be prone to errors.

PR #212, opened 2025-04-12
Add fixing generics for `DeserializeBytes` derive macro

alexeuler:
This PR fixes the issue #88 by adding a custom attribute `eval_generics` for the derive macro. 

Example usage:

```rust
#[derive(Default, Debug, Clone, SerializeBytes, DeserializeBytes)]
#[deserialize_bytes(eval_generics(F = BinaryField128b))]
pub struct MultilinearOracleSet<F: TowerField> {
	oracles: Vec<MultilinearPolyOracle<F>>,
}
```

Will expand the code as follows:

```rust
impl binius_utils::DeserializeBytes for MultilinearOracleSet<BinaryField128b> {
	fn deserialize(
		mut read_buf: impl binius_utils::bytes::Buf,
		mode: binius_utils::SerializationMode,
	) -> Result<Self, binius_utils::SerializationError>
	where
		Self: Sized,
	{
		Ok(Self {
			oracles: binius_utils::DeserializeBytes::deserialize(&mut read_buf, mode)?,
		})
	}
}
```

I took a different approach from what is described in the issue for the following reasons:
1. Overall this feels like more ergonomic (by keeping the number of traits minimal) and generic approach
2. It avoids code duplication for DeserializeBytes derive


jimpo:
Thanks for the PR!! @bergkvist will revie

bergkvist (approved):
This is awesome! Very nice ergonomic improvement, and I like your solution better than what I suggested in the issue

alexeuler:
> This is awesome! Very nice ergonomic improvement, and I like your solution better than what I suggested in the issue

Thanks, @bergkvist! I'm very happy that you loved it :)

jimpo (approved):

PR #214, opened 2025-04-13
Implement SerializeBytes and DeserializeBytes for PackedPrimitiveType…

waamm:
… and ScaledPackedField


jimpo:
This seems pretty straightforward. Can you explain what you need it for? Typically we just serialize scalars instead of packed field elements.

waamm:
> This seems pretty straightforward. Can you explain what you need it for? Typically we just serialize scalars instead of packed field elements.

The underlying goal is to efficiently serialise packed small (say 2-bit) field elements, as serialising millions of these scalars naively would result in significant wasted space. (Currently, I need this functionality to Shamir secret-share and send ZKP witnesses, when all the witness values are binary.)

GraDKh:
The ability to serialize elements in packed may seem useful, but there are many caveats. Our `SerializeBytes` and `SerializeBytes` traits are supposed to support serialization/deserialization across different platforms. So if we ever implement it for some packed field we need to ensure that:
 - It is implemented for all possible packed fields for all platforms. In this particular PR we see that for many underliers the implementation is missing.
 - If for different targets the internal representation of the packed field is different (e.g. `PackedBinaryField4x128b` can either have an M512 underlier if AVX512 is available or be represented as `ScaledPackedField<ScaledPackedField<PackedBinaryField1x128b, 2>>` if only 128b underlier is available natively). I'm not sure this property holds with this.

jimpo:
I agree with @GraDKh 's concern that there may be some cross-platform edge cases that aren't handled correctly, at least at first. Even so, I think the use case outlined is valid so if this works for @waamm I'll merge (once it's passing CI checks).

waamm:
Thanks! This is very helpful feedback. I’ll have some time over the next few days to take a closer look. In the meantime:

> for all possible packed fields

I agree that ideally the `SerializeBytes` and `DeserializeBytes` traits should be added to `PackedField` itself (it would've meant more work than I had the capacity to do then).

> If for different targets the internal representation of the packed field is different

... then the serialisations should be identical - I agree.

> I'm not sure this property holds with this.

I haven't thought this through properly but I would expect that it does - I'll try to write some tests to confirm this.

By the way, why would (for certain architectures) `ScaledPackedField<ScaledPackedField<PackedBinaryField1x128b, 2>, 2>` be preferable to `ScaledPackedField<PackedBinaryField1x128b, 4>`?

GraDKh:
> By the way, why would (for certain architectures) ScaledPackedField<ScaledPackedField<PackedBinaryField1x128b, 2>, 2> be preferable to ScaledPackedField<PackedBinaryField1x128b, 4>?

Those are identical, but `ScaledPackedField<PackedBinaryField2x128b, 2>` maybe be different in case when 256-bit SIMD registers are available (which is actually the case of x64 with AVX2 support).  The idea is that `PackedBinaryField4x128b` may have different representation depending on the target platform:
 - `ScaledPackedField<ScaledPackedField<PackedBinaryField1x128b, 2>, 2>` if 128-bit underlier is supported natively,
 - `ScaledPackedField<PackedBinaryField2x128b, 2>` if 256 SIMD underlier is supported, but not 512-bit
 - PackedBinaryField4x128b, if 512-bit underlier is supported (x64 with AVX512)

So one can use the same type, but arithmetic operations will be performed with the native type that gives the optimal throughput

waamm:
Ah I think the reason `<... , 2>, 2>` appears instead of `<... , 4>` is that you're setting these larger packed fields up "recursively", depending on the architecture.

> I'm not sure this property holds with this.

Could you please take a look at the test I just added to `crates/field/src/arch/portable/packed_scaled.rs` and let me know if it seems convincing?

Also, would you mind running the CI checks again? I haven't had a chance to verify whether everything compiles properly, as I'm guessing emulating AVX instructions on my ARM Mac might present some challenges.

Nitpicky question, coming from a novice: Since AVX appears to treat `__m256i` as `[i64; 4]`, would it potentially be better implement and use (at least in the context of serialization) `impl From<M256> for [i64; 4]` instead of `impl From<M256> for [u128; 2]`?

crates/field/src/arch/portable/byte_sliced/underlier.rs
```diff
@@ -30,6 +34,57 @@ unsafe impl<U: Zeroable, const N: usize> Zeroable for ByteSlicedUnderlier<U, N>
...
+	fn deserialize(
+		read_buf: impl Buf,
+		mode: SerializationMode,
+	) -> Result<Self, SerializationError> {
+		let inner = ScaledUnderlier::<U, N>::deserialize(read_buf, mode)?;
+		Ok(Self(inner))
+	}
+}
+
 impl<U: UnderlierType + Pod, const N: usize> UnderlierType for ByteSlicedUnderlier<U, N> {
```
GraDKh, line 63: Since byte sliced fields are implemented differently than `PackedPrimitiveType<U, Scalar>` this code will never actually be called and byte sliced fields can't actually be byte-serialized at the moment. Which is OK, we can add that support later, but this code won't be used anyway because most probably we would like to serialize them differently (converting to sequential byte order and back).

I see that you had to add this implementation because `UnderlierType` now implies `SerializeBytes` and `DeserializeBytes`. But this is an example of the type that implements `UnderlierType` but doesn't have to implement `SerializeBytes` and `DeserializeBytes`, I believe same for `ScaledUnderlier`, we should have serialization/deserialization implementations on the `PackedField` level. So it seems that `UnderlierType` should not inherit `SerializeBytes` and `DeserializeBytes`, but instead you need to add additional bounds in the implementation for the `PackedPrimitiveType`. And if we need to switch to macros there even additional bounds will be unnecessary as for every particular packed field type we will have some particular underlier.

GraDKh:
> Ah I think the reason <... , 2>, 2> appears instead of <... , 4> is that you're setting these larger packed fields up "recursively", depending on the architecture.

We actually never do it explicitly, just define big packed portable implementations always as `ScaledPackedField<TwiceSmallerPackedField, 2>` to pick up optimal ones for all the architectures. For aarch64 512-bit packed fields become actually `ScaledPackedField<ScaledPackedField<..., 2>, 2>` with underlier `ScaledUnderlier<ScaledUnderlier, 2>, 2>`.

@waamm Unfortunately, I see that it less trivial task than I thought initially. `SerializeBytes` was introduced for a specific reason to allow use different types on seriaization and deserialization sides. To support that there is a special mode to apply the isomorphism to a canonical tower. For example the prover can do all the operations in AES tower and serialize commitment which will be deserialized by a verifier which uses canonical tower only. So to support that we need to apply the isomorphism when serializing/deserializnig packed fields in case when canonical mode is set.

Another moment here is that another natural property for packed fields would be that desalinizing with a different packing width (as long as data size is divisible by that) must give the same result. As long as I see this property holds, but we need some tests for that.

crates/field/src/arch/portable/byte_sliced/underlier.rs
waamm: Ok thanks, I've reverted this.

crates/field/src/arch/portable/packed.rs
```diff
@@ -62,6 +69,31 @@ impl<U: UnderlierType, Scalar: BinaryField> PackedPrimitiveType<U, Scalar> {
...
 
+impl<U: UnderlierType, Scalar: BinaryField> SerializeBytes
+	for PackedPrimitiveType<U, Scalar>
+{
+	fn serialize(
+		&self,
+		write_buf: impl BufMut,
+		mode: SerializationMode,
+	) -> Result<(), SerializationError> {
+		self.0.serialize(write_buf, mode)
```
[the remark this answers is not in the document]
waamm: > But I really doubt that we can do that in a generic form, possible macros are our resort in this case :(

I’ve just started exploring this approach by adding a macro to `crates/field/src/arch/portable/packed.rs`. What do you think?

waamm:
> another natural property for packed fields would be that desalinizing with a different packing width (as long as data size is divisible by that) must give the same result

Did you mean serializing rather than deserializing? Or deserializing followed by something like the `iter` method? I'm not sure I follow otherwise.

GraDKh:
> Did you mean serializing rather than deserializing? Or deserializing followed by something like the iter method? I'm not sure I follow otherwise.

I meant the ability of serialize and deserialize with a different packing width, e.g. serialize `[PackedBinaryField1x128b; 2]` and deserialize it as `[PackedBinaryField2x128b; 1]`

waamm:
> > Did you mean serializing rather than deserializing? Or deserializing followed by something like the iter method? I'm not sure I follow otherwise.
> 
> I meant the ability of serialize and deserialize with a different packing width, e.g. serialize `[PackedBinaryField1x128b; 2]` and deserialize it as `[PackedBinaryField2x128b; 1]`

Sure, and then you can compare the results with either the method `iter()` or the function `pack_slice()` right? I just added a test for this called `test_serialize_deserialize_different_packing_width` in `field/src/packed_binary_field.rs`.

crates/field/src/arch/portable/packed_64.rs
```diff
@@ -16,52 +24,121 @@ use crate::{
...
+// 		impl_mul_with!(PackedBinaryField4x16b => crate::PackedBinaryField8x16b);
+// 		impl_mul_with!(PackedBinaryField2x32b => crate::PackedBinaryField4x32b);
+// 		impl_mul_with!(PackedBinaryField1x64b => crate::PackedBinaryField2x64b);
+// 	} else {
+// 		impl_mul_with!(PackedBinaryField8x8b @ crate::arch::PairwiseTableStrategy);
+// 		impl_mul_with!(PackedBinaryField4x16b @ PairwiseRecursiveStrategy);
+// 		impl_mul_with!(PackedBinaryField2x32b @ PairwiseRecursiveStrategy);
+// 		impl_mul_with!(PackedBinaryField1x64b @ PairwiseRecursiveStrategy);
+// 	}
+// }
```
GraDKh, line 79: Commented code

crates/field/src/arch/portable/packed_64.rs
```diff
@@ -16,52 +24,121 @@ use crate::{
...
+define_packed_field!(
+    PackedBinaryField64x1b, BinaryField1b, u64, 0, (None);
+    PackedBinaryField32x2b, BinaryField2b, u64, 1, (PackedStrategy);
+    PackedBinaryField16x4b, BinaryField4b, u64, 2, (PackedStrategy);
+    PackedBinaryField8x8b,  BinaryField8b,  u64, 3, (CfgSwitch, PackedBinaryField16x8b, PairwiseTableStrategy);
+    PackedBinaryField4x16b, BinaryField16b, u64, 4, (CfgSwitch, PackedBinaryField8x16b, PairwiseRecursiveStrategy);
+    PackedBinaryField2x32b, BinaryField32b, u64, 5, (CfgSwitch, PackedBinaryField4x32b, PairwiseRecursiveStrategy);
+    PackedBinaryField1x64b, BinaryField64b, u64, 6, (CfgSwitch, PackedBinaryField2x64b, PairwiseRecursiveStrategy);
+);
+
```
GraDKh, line 141: Seems like this needs to be reversed

crates/field/src/arch/portable/packed.rs
```diff
@@ -456,6 +460,31 @@ macro_rules! impl_ops_for_zero_height {
 
 pub(crate) use impl_ops_for_zero_height;
 
+macro_rules! serialize_deserialize {
```
GraDKh, line 463: This macro is correct for the canonical tower only, so let's rename to smth like `impl_serialize_deserialize_for_packed_canonical`

crates/field/src/arch/portable/packed.rs
```diff
@@ -36,7 +38,9 @@ pub struct PackedPrimitiveType<U: UnderlierType, Scalar: BinaryField>(
 	pub PhantomData<Scalar>,
 );
 
-impl<U: UnderlierType, Scalar: BinaryField> PackedPrimitiveType<U, Scalar> {
+impl<U: UnderlierType + SerializeBytes + DeserializeBytes, Scalar: BinaryField>
+	PackedPrimitiveType<U, Scalar>
```
GraDKh, line 42: Seems like it can be reversed, `SerializeBytes + DeserializeBytes` bound isn't actually needed

GraDKh (changes requested):
In general it seems good to me, great job! Just several comments left.

There are still AES packed fields and byte sliced fields left, but no need to support them within this PR (and I would actually prefer not to bloat the current one)

crates/field/src/arch/portable/packed.rs
waamm: Yes, thanks.

crates/field/src/arch/portable/packed_64.rs
waamm: I was actually thinking of doing the opposite — expanding `define_packed_field!` and using it across all of the `packed_*.rs` files. This would simplify the code quite a bit, but it's up to you.

GraDKh: As for me that's a bit controversial change. In certain cases we use different strategies for different platforms using `cfg_if` macro.
Anyway better make such change with a separate PR and anyway this macro should be defined not in this file

crates/field/src/arch/portable/packed.rs
```diff
@@ -456,6 +456,31 @@ macro_rules! impl_ops_for_zero_height {
...
 
+macro_rules! impl_serialize_deserialize_for_packed_canonical {
+	($bin_type:ty) => {
+		impl SerializeBytes for $bin_type {
+			fn serialize(
+				&self,
+				write_buf: impl BufMut,
+				mode: SerializationMode,
+			) -> Result<(), SerializationError> {
+				self.0.serialize(write_buf, mode)
```
GraDKh, line 467: A minor suggestion, please add 
```
debug_assert_eq(std::any::TypeId::of::<$bin_type>(), std::any::TypeId::of::<<$bin_type as crate::BinaryField>::Canonical>())
```
to ensure that this macro is actually used for canonical tower only. 

Same for `deserialize`

GraDKh (changes requested):
Sorry for a late review. Just one minor comment and I just realized that there are no implementations for `x86_64/packed_xxx.rs` it seems that it is just a matter of adding macro invocations there. Would be good to have it for consistency. Once it's fixed it's good to merge

crates/field/src/arch/portable/packed.rs
waamm: Right now it is not used for towers and regardless, the trait `crate::BinaryField` has no associated type `Canonical`. Perhaps rename this macro to `impl_serialize_deserialize_for_packed_binary_field`?

GraDKh: Sorry, my bad. Please replace `BinaryField` with `TowerField` and it should work

waamm: In your `debug_assert_eq!` line? It does not, e.g.:

`the trait bound `PackedPrimitiveType<u16, binary_field::BinaryField1b>: binary_field::BinaryField` is not satisfied`
`the trait bound `PackedPrimitiveType<u16, binary_field::BinaryField1b>: binary_field::TowerField` is not satisfied`

GraDKh: Yeah, I missed the fact that this is a packed field. I've checked this line in your branch and it seems to work:

```
debug_assert_eq!(std::any::TypeId::of::<$bin_type>(), std::any::TypeId::of::<<<$bin_type as crate::PackedField>::Scalar as crate::TowerField>::Canonical>());
```

GraDKh (approved):

crates/field/src/arch/portable/packed.rs
waamm: It compiles but the tests are failing for me. Perhaps what you meant is this?

`debug_assert_eq!(std::any::TypeId::of::<<$bin_type as crate::PackedField>::Scalar>(), std::any::TypeId::of::<<<$bin_type as crate::PackedField>::Scalar as crate::TowerField>::Canonical>());`

GraDKh: Yes, this one right, sorry for confusing you

GraDKh (approved):

GraDKh (approved):

jimpo (approved):
Echoing @GraDKh 's review, but as codeowner

waamm:
I hope the `clippy` workflow now finishes; unfortunately I can't run it myself at the moment: `Perfetto tracing is only supported on Linux`.

GraDKh (approved):

jimpo:
Thanks @waamm !

PR #215, opened 2025-04-14
Make projection visible

LindaGuiga:
This PR aims at making `projected` visible, so we can use it for arithmetization.


jimpo (approved):

jimpo (commented):
Sorry, I merged to fast, I'd actually prefer a different interface. Instead of `add_projected` which is very general and not very intuitive, I suggest

```rust
pub fn add_selected_block<FSub, const VALUES_PER_ROW: usize, const NEW_VALUES_PER_ROW: usize>(
		&mut self,
		name: impl ToString,
		col: Col<FSub, VALUES_PER_ROW>,
		index: usize,
	) -> Col<FSub, NEW_VALUES_PER_ROW>
	where
		FSub: TowerField,
		F: ExtensionField<FSub>, { todo!() }
```

This would select the `index`'th block of `NEW_VALUES_PER_ROW` values. This should cover the cases we care about with easier to understand parameters.

PR #217, opened 2025-04-14
Add tracing for unbalanced flushes

Nashtare:
We noticed it was a bit tedious to debug unbalanced channels as the returned error isn't so descriptive (only the channel ID is mentioned).

I added some tracing to get the unbalanced flushes displayed at the `DEBUG` level.

### DIFF
--- a/crates/core/src/constraint_system/channel.rs
+++ b/crates/core/src/constraint_system/channel.rs
@@ -212,6 +212,14 @@ where
 
 	for (id, channel) in channels.iter().enumerate() {
 		if !channel.is_balanced() {
+			let unbalanced_flushes: Vec<_> = channel
+				.multiplicities
+				.iter()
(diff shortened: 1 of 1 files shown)


Nashtare:
On an unrelated note, it may also be useful to add a `fn name(&self) -> &'static str;` method to both channels and tables, as reading an ID when debugging isn't really human-friendly, especially with many tables.

Happy to make a PR if that sounds useful to you too.

jimpo:
> On an unrelated note, it may also be useful to add a `fn name(&self) -> &'static str;` method to both channels and tables, as reading an ID when debugging isn't really human-friendly, especially with many tables.
> 
> Happy to make a PR if that sounds useful to you too.

Hmm, which structs are you referring to exactly? `binius_m3::Table` and `binius_m3::Channel` both have `pub name` fields.

PR #219, opened 2025-04-15
Fix `merge_multilins` for the case when there is a MLE with small number of variables

GraDKh:
This doesn't solve the problem of small tables completely, at least I still wasn't able to prove 16 Groestl permutations. But this is one of the places that needs to be fixed.

Also I see that in several places we needed pretty similar abstractions for the collection of scalars, so I've introduced it in the `binius_field` crate.


crates/field/src/scalars_collection.rs
```diff
@@ -0,0 +1,366 @@
...
+
+use crate::{
+	packed::{get_packed_slice_unchecked, len_packed_slice, set_packed_slice_unchecked},
+	PackedField,
+};
+
+/// A trait for collections of scalars, allowing for indexed access.
+/// This trait is used to abstract over different types of collections - scalar slices,
+/// slices of packed field elements including subranges of collections.
+pub trait ScalarsCollection<F> {
```
jimpo, line 11: I think this should actually go in `binius_util` -- the trait definition isn't specific to fields. I'd call it something more generic like

```rust
pub trait RandomAccessSequence<T: Copy>` {}
```

jimpo (changes requested):

crates/field/src/scalars_collection.rs
GraDKh: Applied

jimpo (approved):

PR #222, opened 2025-04-15
[circuits] Optimised Blake3 compression

storojs72:
This PR re-implements Blake3 compression circuit more optimally for verifier using "horizontal" witness population (inspired by keccakf), which allows proving multiple compressions at once. 

There are couple of questions to reviewers:

1) How should I constrain the `state_i_8` [column](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/blake3.rs#L475) properly? Initially, I had a following issue (when running the entire prove / verify example):
```
PIOPCompilerError(Math(PiecewiseMultilinearIncompatibleEvals { actual: 88, expected: 91 }))

or

Error: PIOP compilation error: binius_math error: there are a total of 50 polynomials, according to n_pieces_by_vars, while you have provided evaluations for 53

Caused by:
    0: binius_math error: there are a total of 50 polynomials, according to n_pieces_by_vars, while you have provided evaluations for 53
    1: there are a total of 50 polynomials, according to n_pieces_by_vars, while you have provided evaluations for 53
```

which at high-level says: "You have some column(s) completely or partially unconstrained". I have identified the problem column - its  `state_i_8`, which is a part of [this](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/blake3.rs#L106) and [this](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/blake3.rs#L106) linear combinations, but I'm not sure what is the problem - so I just left "technical" [constrain](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/blake3.rs#L475): `x - x`. Not sure if this is correct.

2) In keccakf circuit, there is a specific selector-based [constrain](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/keccakf.rs#L430), which I don't understand actually, and I don't know if I need something similar in Blake3. Let me know please.

3) I experimented with [SeveralU32Add](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/lasso/u32add.rs#L49) gadget from lasso lookup package for constraining additions, and it turned out that it is slower than using [regular](https://github.com/storojs72/binius/blob/as/blake3-horizontal/crates/circuits/src/arithmetic/u32.rs#L88) one. Could You please elaborate, what is the benefit of SeveralU32Add and under what conditions it might be better?

### DIFF
--- a/crates/circuits/src/blake3.rs
+++ b/crates/circuits/src/blake3.rs
@@ -1,598 +1,651 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::array;
-
+use anyhow::anyhow;
 use binius_core::oracle::{OracleId, ShiftVariant};
-use binius_field::{BinaryField1b, Field};
+use binius_field::{BinaryField1b, BinaryField32b, Field, TowerField};
+use binius_macros::arith_expr;
+use binius_utils::checked_arithmetics::log2_ceil_usize;
 
 use crate::{
-	arithmetic::{
-		self,
-		u32::{u32const_repeating, LOG_U32_BITS},
-		Flags,
-	},
-	bitwise,
+	arithmetic::u32::LOG_U32_BITS,
 	builder::{types::F, ConstraintSystemBuilder},
 };
 
+const STATE_SIZE: usize = 32;
+
+// This defines how long state columns should be
+const SINGLE_COMPRESSION_N_VARS: usize = 6;
+
+// Number of initial state mutations until getting output value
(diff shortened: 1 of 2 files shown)


jimpo:
@storojs72 Thanks! 

1. Can you check if that error was fixed by https://github.com/IrreducibleOSS/binius/pull/221?

Issue #123, opened 2025-03-14
[arith_expr] Linear-time algorithm for ArithExpr::linear_normal_form


ghost:
The method [`ArithExpr::linear_normal_form`](https://github.com/IrreducibleOSS/binius/blob/5c42cad96db4b2a54c83d24144d5fe7091effe64/crates/math/src/arith_expr.rs#L238) is currently implemented with an easy but inefficient algorithm that is quadratic-time in the number of variables in the worst case. Implement a single-pass version of the same function.

junkicide:
Could you explain what you mean? There's only one loop, so I don't get how it would be quadratic-time even in the worst case.

PR #223, opened 2025-04-16
optimize linear normal form function

junkicide:
solves #123 

instead of initializing a vector each time, an O(n) operation, simply swap elements, an O(1) operation.


jimpo:
Thanks for the attempt. This only solves part of the issue though -- the bigger issue is that `evaluate` is a linear-time algorithm that is called once per variable. So the evaluation algorithm is still quadratic.

The issue was fixed in https://github.com/IrreducibleOSS/binius/pull/224, so I'm going to close this.

PR #224, opened 2025-04-16
Fast (single pass) algorithm for `ArithExpr::linear_normal_form`

alexeuler:
This PR addresses the issue #123.

The idea is to introduce a `SparseLinearNormalForm`, which enables efficient evaluation of an `ArithExpr` under the assumption that the expression is linear.


jimpo (approved):
Thanks @alexeuler!

PR #225, opened 2025-04-16
always sample 4 bytes during `sample_bits`

benediamond:
this is necessary for compatibility with SML.

_as currently written_, the method silently replaces the input argument `bits` with min(`bits`, 64).

the effect of this change is to moreover silently reduce this to 32.

revisit whether there is a more appropriate way to do this?


crates/core/src/transcript/mod.rs
```diff
@@ -434,7 +434,7 @@ where
 fn sample_bits_reader<Reader: Buf>(mut reader: Reader, bits: usize) -> usize {
```
jimpo, line 434: It'd be better to now just return `-> u32` and adjust the line below accordingly.

benediamond: this leads to a big propagation of type changes; are you ok with that?

benediamond: can you take a look at dbe9836? i opted to convert it back to `usize` after the return. we could further redo things so that all of `log_dim`, `fold_arities`, etc., were all `u32`s, instead of `usize`s; but this leads to a pretty big propagation of type changes? let me know your thoughts

jimpo: LGTM now

jimpo (approved):

PR #227, opened 2025-04-16
[piop] Handle arbitrarily small committed polynomials

jimpo:
This relaxes the `ring_switch` and `piop` modules to handle committed polynomials that are smaller than the size of a packed large element and eliminates the `OracleTooSmall` error. We handle this case by committing the polynomial as it's MLE over the larger hypercube, which corresponds to repeating it's evaluations. The integration test in `ring_switch` module is updated to exercise this code path.

### DIFF
--- a/crates/core/src/ring_switch/common.rs
+++ b/crates/core/src/ring_switch/common.rs
@@ -166,16 +166,23 @@ fn group_claims_by_eval_point<F: TowerField>(
 			}
 		})?;
 
-		let (prefix, suffix) = claim.eval_point.split_at(kappa);
+		let (prefix, suffix) = if claim.eval_point.len() < kappa {
+			// If evaluation point is less than kappa, pad the evaluation point with 0s
+			let mut prefix = Vec::with_capacity(kappa);
+			prefix.extend_from_slice(&claim.eval_point);
+			prefix.resize(kappa, F::ZERO);
+			(prefix, &[][..])
+		} else {
+			let (prefix, suffix) = claim.eval_point.split_at(kappa);
+			(prefix.to_vec(), suffix)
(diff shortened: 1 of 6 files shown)


crates/core/src/ring_switch/common.rs
```diff
@@ -166,16 +166,23 @@ fn group_claims_by_eval_point<F: TowerField>(
 			}
 		})?;
 
-		let (prefix, suffix) = claim.eval_point.split_at(kappa);
+		let (prefix, suffix) = if claim.eval_point.len() < kappa {
+			// If evaluation point is less than kappa, pad the evaluation point with 0s
+			let mut prefix = Vec::with_capacity(kappa);
+			prefix.extend_from_slice(&claim.eval_point);
+			prefix.resize(kappa, F::ZERO);
```
josephjohnston, line 173: do you really have to pad with zeros? if you evaluate with the whole evaluation point, rather than replacing the extra vars with zeros, I think you should get the same evaluation (because this is Lagrange basis)

josephjohnston: must we pad with zeros because only the _honest_ prover necessarily repeats the evaluations (rather than filling with non-repeating evaluations), and the verifier can't enforce this to be the case?

jimpo: You're right, any extension of the evaluation point would be secure to use here. This is just the easiest one to choose and there's not too much of a penalty choosing this versus another one. The best choice would be another longer prefix used for another polynomial, so the prover would send one fewer tensor algebra element. It's just more code to figure out what that point is.

josephjohnston (approved):

PR #228, opened 2025-04-16
SYS-110: compute API supports host buffer management, allocation

BenjaminTrapani:
Accelerated compute backends typically require setting up IOMMU or performing custom allocation and alignment for buffers that are copyable to and from device memory. Extend the `ComputeMemory` trait with a type that is transparent to Binius as a dense slice of elements with a minimum slice width.

Define a `ComputeBufferAllocator` trait supporting allocating host and device memory. Implementations will provide specializations of `ComputeMemory` and `ComputeBufferAllocator` that perform the device-specific setup for both classes of allocations in addition to specifying the implementation-specific constraints on minimum span widths.

### DIFF
--- a/crates/compute/src/alloc.rs
+++ b/crates/compute/src/alloc.rs
@@ -2,75 +2,35 @@
 
 use std::cell::Cell;
 
-use crate::memory::{ComputeMemoryOperations, ComputeMemoryTypes, OpaqueSlice};
-/// A trait for allocating slices from a compute memory buffer.
-///
-/// This trait provides an interface for allocating fixed-size slices from an underlying
-/// memory buffer, such as device memory or host memory. The allocator maintains ownership
-/// of the buffer and hands out slices with lifetimes tied to the allocator.
-///
-/// # Type Parameters
-///
-/// - `'a`: The lifetime of the allocator and allocated slices
-/// - `F`: The element type stored in the buffer
-/// - `Mem`: The memory types trait defining slice representations
-/// - `MemOperations`: The memory operations trait for manipulating slices
-///
-/// # Safety
-///
-/// Implementations must ensure:
-/// - Allocated slices do not overlap
-/// - Slice lengths are valid multiples of `Mem::MIN_SLICE_LEN`
-/// - Slices remain valid for their full lifetime
-/// - The total allocation size does not exceed the underlying buffer
-pub trait ComputeBufferAllocator<'a, F, Mem, MemOperations>
-where
-	Mem: ComputeMemoryTypes<F>,
-	MemOperations: ComputeMemoryOperations<F, Mem>,
-{
-	/// Allocates a slice of elements.
-	///
-	/// This method operates on an immutable self reference so that multiple allocator references
-	/// can co-exist.
-	///
-	/// ## Pre-conditions
-	///
-	/// - `n` must be a multiple of `Mem::MIN_SLICE_LEN`
-	fn alloc(&self, n: usize) -> Result<Mem::FSliceMut<'a>, Error>;
-}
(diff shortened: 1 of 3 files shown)


jimpo (changes requested):
There are a few changes in here:

1. *Have the compute layer expose a method for getting a host buffer allocator*. Fully agree we need this.
2. *Make host memory type-safe*. I have discussed the merits of this versus a less type-safe approach of just using regular slices with the SysSW team before. The alternative is that methods that use host memory take regular `&[F]` slices, and it's a function precondition for memory transfer methods that this was allocated by the HAL's host memory allocator. This reduces the amount of code while giving up on some static analysis. I think @alsrdn and I agreed on that, but we could revisit the decision.
3. *Abstract allocation behind a trait*. We could, but I also think directly returning a `BumpAllocator` struct from `alloc_host` and `alloc_device` will serve us perfectly well for now.
4. *Have ComputeMemory know about both device and host memory*. This I don't like as much. The alternative is that the `ComputeLayer` trait, which will be the main object that the code has access to, could have two `ComputeMemory` associated traits (instead of inheriting from `ComputeMemory`). This is sort of a type-level composition vs inheritance question. Something like

```rust
trait ComputeLayer<F> {
  type DevMem: ComputeMemory<F>;
  type HostMem: ComputeMemory<F>
}
```

(`ComputeLayer` is still in design [here](https://github.com/IrreducibleOSS/binius/blob/jimpo/2024-04-13-hal-parallel-ops/crates/compute/src/layer.rs))

The other tricky thing is that `HostMem` needs `FSlice` and `FSliceMut` to implement `AsRef<[F]>` and `AsMut<[F]>` for accessing memory. It could be kind of tricky to enforce those extra type bounds, I haven't tried yet.

My recommendation overall would be to have `ComputeLayer` have two methods:

```rust
trait ComputeLayer<F>: ComputeMemory<F> {
    fn dev_alloc(&self) -> BumpAllocator<'_, F, Self>;
    fn host_alloc(&self) -> BumpAllocator<'_, F, CpuMemory>;
}
```

BenjaminTrapani:
Thank you for the PR comments! Based on this feedback, I propose the following changes:

1. Split the `ComputeMemory` trait into `ComputeMemoryDevice` and `ComputeMemoryHost`. `ComputeMemoryDevice` constrains `FSlice` to `DevSlice` and `ComputeMemoryHost` constrains with `AsRef<F>`. Letting the host memory implementation use a type besides a raw slice lets the implementation include extra information required for host<->device transfers in the host handle (iova for example).
2. Update the allocator trait and implementation to have one `alloc` function and one `reset` function, letting Binius continue to use a bump-allocator-like interface for all implementations.
3. Expose two allocator types in `ComputeLayer`, one for host allocations and another for device allocations.
4. Provide functions in `ComputeLayer` supporting transferring data between host and device handles.

Providing an allocator trait allows implementations to perform device-specific setup for host buffers on allocation, saving an allocation of a temporary buffer prior to the transfer to device as noted in https://github.com/IrreducibleOSS/binius/pull/228#issuecomment-2812332583. For implementations that use IOMMU, an implementation-defined allocator enables setting up the IOMMU tables incrementally, improving performance over a variant that mmaps a large block on program startup and initializes the IOMMU table for the entire preallocated block.

jimpo:
Thanks for the updates @BenjaminTrapani. This still looks like it has many new concepts, abstractions, and type bounds that we can avoid with a simpler approach. I've prototyped that in https://github.com/IrreducibleOSS/binius/pull/240 and I think it should work reasonably well.

BenjaminTrapani:
Closing in favor of https://github.com/IrreducibleOSS/binius/pull/248 and https://github.com/IrreducibleOSS/binius/pull/240.

PR #231, opened 2025-04-17
Fix shape of terminal FRI oracle

djadjka:
Update the batch_size of the `terminate_codeword` Merkle tree so that it has exactly `inv_rate` leaves

### DIFF
--- a/crates/core/src/protocols/fri/prove.rs
+++ b/crates/core/src/protocols/fri/prove.rs
@@ -381,12 +381,6 @@ where
 			.get(self.round_committed.len() + 1)
 			.map(|log| 1 << log)
 			.unwrap_or_else(|| {
-				println!(
-					"fold {}",
-					log2_strict_usize(folded_codeword.len())
-						.saturating_sub(self.params.rs_code().log_inv_rate())
-				);
-
 				1 << log2_strict_usize(folded_codeword.len())
 					.saturating_sub(self.params.rs_code().log_inv_rate())
 			});
--- a/crates/core/src/protocols/fri/prove.rs
+++ b/crates/core/src/protocols/fri/prove.rs
@@ -6,7 +6,7 @@ use binius_field::{
 };
 use binius_hal::{make_portable_backend, ComputationBackend};
 use binius_maybe_rayon::prelude::*;
-use binius_utils::{bail, SerializeBytes};
+use binius_utils::{bail, checked_arithmetics::log2_strict_usize, SerializeBytes};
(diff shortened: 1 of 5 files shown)


crates/core/src/protocols/fri/prove.rs
```diff
@@ -231,12 +231,10 @@ where
...
-	// take the first arity as coset_log_len, or use log_inv_rate if arities are empty
-	let coset_log_len = params
-		.fold_arities()
-		.first()
-		.copied()
-		.unwrap_or_else(|| rs_code.log_inv_rate());
+	// Take the first arity as coset_log_len, or use the value such that the number of leaves equals 1 << log_inv_rate if arities is empty
+	let coset_log_len = params.fold_arities().first().copied().unwrap_or_else(|| {
+		log2_strict_usize(encoded.len() * P::WIDTH).saturating_sub(rs_code.log_inv_rate())
+	});
```
benediamond, line 237:
```suggestion
	let coset_log_len = params.fold_arities().first().copied().unwrap_or_else(|| {
		log_elems
	});
```
this can be simplified.

i think you could also get rid of the other `log2_strict_usize` below, but it's slightly less obvious how

crates/core/src/protocols/fri/prove.rs
```diff
@@ -382,7 +380,10 @@ where
 			.fold_arities()
 			.get(self.round_committed.len() + 1)
 			.map(|log| 1 << log)
-			.unwrap_or_else(|| self.params.rs_code().inv_rate());
+			.unwrap_or_else(|| {
+				1 << log2_strict_usize(folded_codeword.len())
+					.saturating_sub(self.params.rs_code().log_inv_rate())
+			});
```
benediamond, line 386:
```suggestion
			.unwrap_or_else(|| {
				1 << self.params.n_final_challenges()
			});
```
indeed, this too can be simplified; and you can get rid of `log2_strict_usize

benediamond:
done with review; please see comments / suggestions above

crates/core/src/protocols/fri/prove.rs
djadjka: oh, thank you!

benediamond:
@djadjka i didn't look at the tests—maybe you can look at those too, and see if they can be simplified in a similar way?

djadjka:
> @djadjka i didn't look at the tests—maybe you can look at those too, and see if they can be simplified in a similar way?

I want to save naive approach in tests

jimpo (approved):

PR #235, opened 2025-04-18
Common expression extraction in `ArithExpr`

GraDKh:
Several changes to support sub-expression deduplication in `ExprTree` and ArithCircuitPoly:
1. Use `Arc` in `ArithExpr`
2. Introduce `ArithCircuit` as an intermediate object during `ArithExpr` serialization/deserialization of `ArithExpr` to preserve the `Arc`s keep pointing the same nodes after serialization roundtrip. This structure is not public.
3. At the moment of creation `ArithCircuitPoly` automatically deduplicates the nodes, reusing already created steps if possible. This seems to be of almost the same code complexity, so IMO there is no reason to introduce a new public entity `ArithExpr`.

### DIFF
--- a/crates/core/src/polynomial/arith_circuit.rs
+++ b/crates/core/src/polynomial/arith_circuit.rs
@@ -1207,7 +1207,7 @@ mod tests {
 
 		let (steps, retval) = convert_circuit(&(&expr).into());
 
-		assert_eq!(steps.len(), 3, "Expression should generate 4 computation steps");
+		assert_eq!(steps.len(), 3, "Expression should generate 3 computation steps");
 
 		assert!(
 			matches!(
--- a/crates/core/src/polynomial/arith_circuit.rs
+++ b/crates/core/src/polynomial/arith_circuit.rs
@@ -1,9 +1,11 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::{collections::HashMap, fmt::Debug, mem::MaybeUninit, sync::Arc};
+use std::{fmt::Debug, mem::MaybeUninit, sync::Arc};
 
 use binius_field::{ExtensionField, Field, PackedField, TowerField};
-use binius_math::{ArithExpr, CompositionPoly, Error, RowsBatchRef};
+use binius_math::{
+	ArithCircuit, ArithCircuitStep, ArithExpr, CompositionPoly, Error, RowsBatchRef,
+};
 use binius_utils::{bail, DeserializeBytes, SerializationError, SerializationMode, SerializeBytes};
 use stackalloc::{
 	helpers::{slice_assume_init, slice_assume_init_mut},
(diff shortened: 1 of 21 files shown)


jimpo (commented):
This is a much larger change than I was expecting. I think this could be done with just a couple changes:

1. Box -> Arc in ArithExpr.
2. Create an `ArithCircuit` struct that serializes the nodes and converts children refs to index references. Deduplicate nodes using a linear scan over the serialized nodes.
3. Create `ArithCircuitPoly` from ArithCircuit instead of ArithExpr.

crates/math/src/arith_expr.rs
```diff
@@ -441,6 +615,226 @@ impl<F: Field> Product for ArithExpr<F> {
...
+				ArithExpr::Var(_) => 0, // variables do not represent a separate step
+				ArithExpr::Add(left, right) | ArithExpr::Mul(left, right) => {
+					1 + estimate_steps(left) + estimate_steps(right)
+				}
+				ArithExpr::Pow(base, _) => 1 + estimate_steps(base),
+			}
+		}
+
+		let steps_bound = estimate_steps(self);
+		stackalloc_with_default::<ArithCircuitStep<F>, _, _>(steps_bound, |mut steps_buf| {
```
jimpo, line 637: Please just use a Vec. This isn't performance critical, and we should only use `stackalloc` where absolutely necessary.

crates/math/src/arith_expr.rs
```diff
@@ -18,13 +23,16 @@ use super::error::Error;
 /// Arithmetic expressions are trees, where the leaves are either constants or variables, and the
 /// non-leaf nodes are arithmetic operations, such as addition, multiplication, etc. They are
 /// specific representations of multivariate polynomials.
-#[derive(Debug, Clone, PartialEq, Eq, SerializeBytes, DeserializeBytes)]
+///
+/// The `Arc`'s are not guaranteed to be unique, so the expression tree may contain duplicate nodes.
+/// Use `deduplicate_nodes` to remove duplicate nodes from the expression tree.
```
jimpo, line 28:
```suggestion
/// Use [`Self::deduplicate_nodes`] to remove duplicate nodes from the expression tree.
```

crates/math/src/arith_expr.rs
```diff
@@ -116,60 +124,112 @@ impl<F: Field> ArithExpr<F> {
 	/// * [`Error::IncorrectArgumentLength`] if indices has length less than the current number of
 	///   variables
 	pub fn remap_vars(self, indices: &[usize]) -> Result<Self, Error> {
```
jimpo, line 126: This is complicated now. We should just move it to `ArithCircuit`, where it will be easier to implement.

That can be a follow-up.

crates/math/src/arith_expr.rs
```diff
@@ -116,60 +124,112 @@ impl<F: Field> ArithExpr<F> {
...
-				Self::Pow(Box::new(new_base), exp)
+				let new_base = remap_vars_inner(&base, indices)?;
+				Self::Pow(new_base, exp)
 			}
 		};
 		Ok(expr)
 	}
 
 	/// Substitute variable with index `var` with a constant `value`
 	pub fn const_subst(self, var: usize, value: F) -> Self {
```
jimpo, line 183: Similar to `remap_vars`

crates/math/src/arith_expr.rs
```diff
@@ -441,6 +615,226 @@ impl<F: Field> Product for ArithExpr<F> {
...
+	Const(F),
+}
+
+impl<F: Field> Default for ArithCircuitStep<F> {
+	fn default() -> Self {
+		Self::Const(F::ZERO)
+	}
+}
+
+/// A simple circuit representation of an arithmetic expression.
```
jimpo, line 675: newline after the first line of docs. See https://github.com/IrreducibleOSS/binius/blob/main/CONTRIBUTING.md#documentation

crates/math/src/arith_expr.rs
```diff
@@ -441,6 +615,226 @@ impl<F: Field> Product for ArithExpr<F> {
...
+/// We use this representation as a temporary object for serialization/deserialization to
+/// ensure that common sub-expressions are correctly desewrialized into the same `Arc` instance.
+#[derive(Clone, Debug, SerializeBytes, DeserializeBytes)]
+struct ArithCircuit<F: Field, Data: AsRef<[ArithCircuitStep<F>]>> {
```
jimpo, line 679: Does this need to be generic over `Data`? `Vec` should be good enough.

crates/math/src/arith_expr.rs
```diff
@@ -441,6 +615,226 @@ impl<F: Field> Product for ArithExpr<F> {
...
+
+		ArithCircuit {
+			steps: &data.as_mut()[..current_len],
+			_pd: PhantomData,
+		}
+	}
+
+	/// Converts the circuit back to an expression.
+	fn to_expr(&self) -> ArithExpr<F> {
+		let root = stackalloc_with_default::<Option<Arc<ArithExpr<F>>>, _, _>(
```
jimpo, line 785: Do this with regular allocation instead of stack allocation

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -10,51 +10,103 @@ use stackalloc::{
 	stackalloc_uninit,
 };
 
+/// Convert the expression to a sequence of arithmetic operations that can be evaluated in sequence.
 /// Convert the expression to a sequence of arithmetic operations that can be evaluated in sequence.
 fn circuit_steps_for_expr<F: Field>(
```
jimpo, line 15: We should change this to convert `ArithCircuit` to `ArithCircuitPoly` instead of converting from `ArithExpr`. `ArithExpr` -> `ArithCircuit` -> `ArithCircuitPoly`.

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -10,51 +10,103 @@ use stackalloc::{
 	stackalloc_uninit,
 };
 
+/// Convert the expression to a sequence of arithmetic operations that can be evaluated in sequence.
```
jimpo, line 13: duplicate line

jimpo (changes requested):

crates/math/src/arith_expr.rs
GraDKh: Within this change we need to have it in `ArithExpr` as it is still in use

jimpo: ArithExpr::remap_vars can call `ArithCircuit::new` -> `remap_vars` -> `ArithCircuit::into` for backwards compatibility.

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -159,22 +216,28 @@ impl<F: TowerField> DeserializeBytes for ArithCircuitPoly<F> {
...
 	}
 }
 
 impl<F: TowerField> ArithCircuitPoly<F> {
-	pub fn new(expr: ArithExpr<F>) -> Self {
-		let degree = expr.degree();
-		let n_vars = expr.n_vars();
-		let tower_level = expr.binary_tower_level();
-		let (exprs, retval) = circuit_steps_for_expr(&expr);
+	pub fn new(expr: &ArithExpr<F>) -> Self {
```
jimpo, line 229: There are a lot of lines in the diff because of this change. No good reason to change it back at this point, but it should be fine to have kept it as it was -- cloning Arcs is cheap.

jimpo (approved):

PR #237, opened 2025-04-18
[m3] fix OOB in Display impl of ConstraintSystem

pepyakin:
If you try to display the constraint system in the m3/collatz test, you would get

```
thread 'arithmetization::test_collatz_validate_witness' panicked at /Users/pepyakin/dev/irreducible/binius/crates/m3/src/builder/constraint_system.rs:58:47:
index out of bounds: the len is 3 but the index is 3
```

I am not 100% sure in the fix, but it seems like `column_indices` are in the `table.columns` index space.


jimpo (approved):
Thanks, looks right to me

PR #239, opened 2025-04-18
[gadgets] Adds unsigned multiplication gadgets to M3

Anex007:
This MR adds unsigned multiplication gadgets to M3, as well as expose GKR exponentiation into `M3`. Currently I've only added `u32` and `u64`


crates/m3/src/builder/table.rs
```diff
@@ -258,6 +258,81 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
...
+		&mut self,
+		name: impl ToString,
+		pow_bits: &[Col<B1>],
+		base: F,
+	) -> Col<FSub>
+	where
+		FSub: TowerField,
+		F: ExtensionField<FSub>,
+	{
+		assert!(pow_bits.len().is_power_of_two());
```
djadjka, line 283: "In fact, gkr_exp works not only with sizes that are powers of two, so I would prefer to remove this restriction to allow for special optimizations.

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,432 @@
...
+		let g_pow_x: Col<FPExt> = table.add_static_exp("g^x", &x_in_bits, generator);
+		let g_pow_xy: Col<FPExt> = table.add_dynamic_exp("(g^x)^y", &y_in_bits, g_pow_x);
+
+		let out_high_bits = array::from_fn(|i| table.add_committed(format!("out_high[{i}]")));
+		let out_low_bits = array::from_fn(|i| table.add_committed(format!("out_low[{i}]")));
+
+		let out_high = table.add_computed("out_high", pack_fp(out_high_bits));
+		let out_low = table.add_computed("out_low", pack_fp(out_low_bits));
+
+		let g_pow_out_low: Col<FPExt> =
```
djadjka, line 162: I think we should use add_static_exp\<FPExt\> instead of Col\<FPExt\>.

crates/m3/src/builder/column.rs
```diff
@@ -152,4 +152,14 @@ pub enum ColumnDef<F: TowerField = B128> {
...
 	},
+	StaticExp {
+		bit_cols: Vec<ColumnIndex>,
+		base: F,
+		base_tower_level: usize,
+	},
+	DynamicExp {
+		bit_cols: Vec<ColumnIndex>,
+		base: ColumnIndex,
+		base_tower_level: usize,
```
djadjka, line 163: We don't need `base_tower_level` here we can receive them from oracles

crates/m3/src/builder/table.rs
```diff
@@ -258,6 +258,81 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
...
+		assert!(pow_bits.len() <= (1 << FSub::TOWER_LEVEL));
+
+		let namespaced_name = self.namespaced_name(name);
+		let bit_cols = pow_bits.iter().map(|bit| bit.id().table_index).collect();
+		self.table.new_column(
+			namespaced_name,
+			ColumnDef::DynamicExp {
+				bit_cols,
+				base: base.id().table_index,
+				base_tower_level: FSub::TOWER_LEVEL,
```
djadjka, line 331: `FSub::TOWER_LEVEL` is very expensive, you should add FExpBase (see old circuit)

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,432 @@
...
+	fn is_bit_set_at(a: B64, index: usize) -> bool {
+		((a.val() >> index) & 1) == 1
+	}
+
+	fn get_generator() -> Self::FPExt {
+		B128::MULTIPLICATIVE_GENERATOR
+	}
+
+	fn get_shifted_generator() -> B128 {
+		let base = Self::get_generator();
```
djadjka, line 83: pls use 

```
fn exp_pow2<F: BinaryField>(mut g: F, log_exp: usize) -> F {
	for _ in 0..log_exp {
		g *= g
	}
	g
}
```

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,432 @@
...
+	fn is_bit_set_at(a: B32, index: usize) -> bool {
+		((a.val() >> index) & 1) == 1
+	}
+
+	fn get_generator() -> B64 {
+		B64::MULTIPLICATIVE_GENERATOR
+	}
+
+	fn get_shifted_generator() -> B64 {
+		B64::MULTIPLICATIVE_GENERATOR.pow(1 << Self::get_bit_length())
```
djadjka, line 55: exp_pow2 to

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,432 @@
...
+		let mut out_high = index.get_mut(self.out_high)?;
+
+		for (i, (x, y)) in x_vals.into_iter().zip(y_vals.into_iter()).enumerate() {
+			let (res_high, res_low) = UX::mul(x, y);
+			set_packed_slice(&mut x_in, i, x);
+			set_packed_slice(&mut y_in, i, y);
+			set_packed_slice(&mut out_low, i, res_low);
+			set_packed_slice(&mut out_high, i, res_high);
+
+			for bit_idx in 0..BIT_LENGTH {
```
djadjka, line 230: Can we try to optimize it, using rayon?
like columns_to_numbers and numbers_to_columns

djadjka (changes requested):
Nice work, I have a few comments.
Could I ask you to update the examples to use m3?
Also, it would be great to check the traces to avoid any unexpected regressions.

crates/m3/src/builder/column.rs
Anex007: It's used to create automatically create committed columns in `constraint_system::add_oracle_for_column`

crates/m3/src/gadgets/mul.rs
Anex007: `(0..BIT_LENGTH).into_par_iter()` results in error: `binius_maybe_rayon::range::Iter<usize>` is not an iterator`
we can possibly optimize this after all the multiplication and division gadgets are Merged in.

Anex007:
> Nice work, I have a few comments. Could I ask you to update the examples to use m3? Also, it would be great to check the traces to avoid any unexpected regressions.

Updated both examples to the new M3 circtuit, parallel trace generation can be handled by a second PR after all the multiplication and division gadgets are merged in.

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,431 @@
...
+		&self,
+		index: &mut TableWitnessSegment<P>,
+		x_vals: impl IntoIterator<Item = FP>,
+		y_vals: impl IntoIterator<Item = FP>,
+		fill_input_bits: bool,
+	) -> Result<()>
+	where
+		P: PackedField<Scalar = B128> + PackedExtension<B1> + PackedExtension<FP>,
+	{
+		let mut x_in_bits = self
```
jimpo, line 196: Can you use [array_util::try_map](https://docs.rs/array-util/latest/array_util/fn.try_map.html) here instead?

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,431 @@
...
+};
+
+use crate::builder::{
+	upcast_col, Col, Expr, TableBuilder, TableWitnessSegment, B1, B128, B32, B64,
+};
+
+trait UnsignedMulPrimitives {
+	type FP: TowerField;
+	type FExpBase: TowerField + ExtensionField<Self::FP>;
+	fn get_bit_length() -> usize;
```
jimpo, line 18: This seems like it could be an associated const? `const BIT_LENGTH: usize`

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,431 @@
...
+use binius_field::{
+	ext_basis, packed::set_packed_slice, BinaryField, ExtensionField, Field, PackedExtension,
+	PackedField, TowerField,
+};
+
+use crate::builder::{
+	upcast_col, Col, Expr, TableBuilder, TableWitnessSegment, B1, B128, B32, B64,
+};
+
+trait UnsignedMulPrimitives {
```
jimpo, line 15: trait docstring please

jimpo (commented):
Nice, looks good overall.

examples/u32_mul_gkr_exp_lookups.rs
```diff
@@ -1,21 +1,27 @@
 // Copyright 2024-2025 Irreducible Inc.
```
jimpo, line 1: This file has "lookups" in the name but doesn't do any lookups?

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,408 @@
...
+		FExpBase: TowerField,
+		FP: TowerField,
+		UX: UnsignedMulPrimitives<FP = FP, FExpBase = FExpBase>,
+		const BIT_LENGTH: usize,
+	> Mul<FExpBase, FP, UX, BIT_LENGTH>
+where
+	FExpBase: ExtensionField<FP> + ExtensionField<B1>,
+	B128: ExtensionField<FExpBase> + ExtensionField<FP> + ExtensionField<B1>,
+{
+	// TODO: Figure out if we need to create a separate table for this gadget.
```
jimpo, line 119: Probably not?

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,408 @@
...
+			g *= g
+		}
+		g
+	}
+}
+
+// Internally used to have deduplicated implementations.
+#[derive(Debug)]
+struct Mul<
+	FExpBase: TowerField,
```
jimpo, line 91: These type parameters are redundant since they are associated types on `UX`. Please remove them from the `struct` type parameters, but you can keep them on the `impl` block for convenience.

crates/m3/src/gadgets/mul.rs
```diff
@@ -0,0 +1,408 @@
...
+			for bit_idx in 0..BIT_LENGTH {
+				if fill_input_bits {
+					if UX::is_bit_set_at(x, bit_idx) {
+						set_packed_slice(&mut x_in_bits[bit_idx], i, B1::ONE);
+					}
+					if UX::is_bit_set_at(y, bit_idx) {
+						set_packed_slice(&mut y_in_bits[bit_idx], i, B1::ONE);
+					}
+				}
+				if UX::is_bit_set_at(res_low, bit_idx) {
```
jimpo, line 215: Instead of the `if`s , just do `set_packed_slice(&mut out_low_bits[bit_idx], i, B1::from(UX::is_bit_set_at(res_low, bit_idx))`

jimpo (changes requested):

jimpo (approved):

crates/m3/src/gadgets/mul.rs
djadjka: Could you please add to-do or create ticket for it, now generation of trace 4x slower

djadjka (approved):
I’d prefer to restore the u32_mul_gkr_exp_lookups example, but that can be done in a separate PR.

examples/u32_mul_gkr_exp_lookups.rs
```diff
@@ -1,99 +0,0 @@
-// Copyright 2024-2025 Irreducible Inc.
```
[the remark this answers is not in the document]
jimpo: @Anex007 Please do this as a follow-up

PR #242, opened 2025-04-21
[m3] u32 sub gadget

pepyakin:
This mostly repeats the shape seen in U32Add.

I took a different approach with tests here which reduces duplication and seems a bit cleaner compared to the status quo.


jimpo (approved):
Awesome, nice job with the test harness!

PR #244, opened 2025-04-21
"Hybrid" zerocheck

onesk:
This PR intends to fix the zerocheck univariate round soundness hole by always skipping the same number of rounds, resulting in same-sized round polynomials in Lagrange basis. Remaining rounds are "right justified" using front-loaded batching, resulting in a peculiar reduction shape where projection happens on the low variables (that underwent "univariate skip") and possibly a suffix of high variables, potentially leaving a "gap" in between. The primary reason for this contrived arrangement is to be able to reuse high projection from univariatizing reduction witness generation in ring switching.

I made a decision to abandon any aspirations to univariate skip sumcheck and make a clear distinction between sumchecks (regular and eq-ind) and zerochecks (which are proven using this complex three-step reduction of univariate skip, front-loaded batched sumcheck, and univariatizing sumcheck). In my opinion this resulted in massive simplification of zerocheck calling contract: where you used to do a series of reductions you now simply `batch_verify_sumcheck/batch_verify_zerocheck`, and inner machinery stays hidden. However the changeset turned out to be quite massive, and I can only hope it does not obscure the intent too much.


jimpo:
Taking over this MR. I'll merge and do a thorough review later.

jimpo:
@milosbackonja Any updates?

PR #250, opened 2025-04-23
Use `ArithExpr` as a builder only. All other operations are now moved to `ArithCircuit`

GraDKh:
To my opinion now it is much more consistent. `ArithExpr`s are used to only build certain parts of tree but never stored.


crates/math/src/arith_expr.rs
```diff
@@ -924,8 +677,81 @@ impl<F: Field> From<SparseLinearNormalForm<F>> for LinearNormalForm<F> {
...
+				self.prev_step(*base) == other.prev_step(*other_base) && exp == other_exp
+			}
+			_ => false,
+		}
+	}
+}
+
+impl<F: Field> Eq for StepNode<'_, F> {}
+
+impl<F: Field> Hash for StepNode<'_, F> {
```
jimpo, line 721: Why do we need this?

crates/math/src/arith_expr.rs
```diff
@@ -727,116 +229,367 @@ impl<F: Field> ArithCircuit<F> {
...
+	fn add(mut self, rhs: Self) -> Self {
+		self += rhs;
+		self
+	}
+}
+
+impl<F: Field> AddAssign for ArithExpr<F> {
+	fn add_assign(&mut self, mut rhs: Self) {
+		let old_len = self.steps.len();
+		add_offset(&mut rhs.steps, old_len);
```
jimpo, line 525: Won't this duplicate nodes if the lhs and rhs are expressions sharing a common subexpression?

crates/math/src/arith_expr.rs
GraDKh: This is a helper struct to compare the subtrees by their values and store in hashed collections. This particular is needed to find the repeating subexpressions in the `deduplicate_steps`

crates/math/src/arith_expr.rs
GraDKh: Yes it will. The workflow is to build the expression and call `optimize` once at the end (it is called at the moment creation of the ArithCircuitPoly). This will extract all the duplicate nodes and compress the circuit

jimpo: This looks like a regression in functionality to me. Please keep ArithExpr, which dedups by construction.

GraDKh: We can automatically call the deduplication after each operation which would give the similar result.

But TBH calling `optimize` before we're constructing some computation object for the expression (`ArithCircuitPoly` or generating instruction for FPGA) is necessary anyway. Because after we've constructed the polynomials manually we can apply arithmetic operations to them in compositions, partially substitute variables with constants, etc. So the best place to make the constant folding/extract common steps is there.

Or maybe I'm missing something and there are some situations when we want to ensure that certain steps will be shared before that?

crates/math/src/arith_expr.rs
```diff
@@ -793,50 +716,140 @@ impl<F: Field> From<&ArithExpr<F>> for ArithCircuit<F> {
...
+		} == StepNode {
+			index: other.steps.len() - 1,
+			steps: &other.steps,
 		}
+	}
+}
 
-		let mut cached_exprs = vec![None; circuit.steps.len()];
-		let root = visit_step(circuit.steps.len() - 1, &mut cached_exprs, &circuit.steps);
+impl<F: Field> Add for ArithCircuit<F> {
```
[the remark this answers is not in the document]
jimpo: Discussed with Dmitry, and we're going to keep these.

crates/m3/src/builder/expr.rs
```diff
@@ -21,7 +21,7 @@ pub struct Expr<F: TowerField, const V: usize> {
 	#[get_copy = "pub"]
 	table_id: TableId,
 	#[get = "pub"]
-	expr: ArithExpr<F>,
+	expr: ArithCircuit<F>,
```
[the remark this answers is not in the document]
GraDKh: I tried applying this change but in the code below we need `degree` and `convert_field` methods, so we need to duplicate them in `ArithExpr`. And TBH I see no reason to have the `ArithExpr` here as when we are manipulating with expressions we never share common sub-expressions from already created ones. So the code with `ArithCircuit` and `ArithExpr` produces equivalent results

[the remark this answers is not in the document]
jimpo: We will need the ability to share subexpressions here.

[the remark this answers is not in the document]
GraDKh: As discussed, switched to the conversion to `ArithCircuit` where needed.

jimpo (dismissed):

jimpo (approved):

PR #253, opened 2025-04-24
ci: investigate oom

fkondej:


.github/workflows/ci.yml
```diff
@@ -105,7 +106,7 @@ jobs:
           - runner: "c7a-2xlarge"
             name: "no-default-features"
             cmd: 'CARGO_EXTRA_FLAGS="--no-default-features" RUSTFLAGS="-C target-cpu=native" ./scripts/run_tests_and_examples.sh'
-          - runner: "c8g-2xlarge"
+          - runner: "c8g-4xlarge"
```
jimpo, line 109: I think switching to a higher mem instance like m8g or r8g will be more likely to help.

milosbackonja: Yea, I agree. it will be more balanced. Actually last run was on r8g.2xlarge
@jimpo does it make sense to disable profiling for those unit tests if its possible to reduce memory footprint?

jimpo: Sure

GraDKh: @learyducible What do you mean by profiling? If tracing then we have tracing enabled when running examples only, not unit tests. To disable collecting tracing events we can switch `RUST_LOG` to `error`

jimpo: @learyducible @milosbackonja Was the profiling suggestion something that you've seen work or just a guess? This isn't a problem with memory consumption during runtime, it's a compile/link-time issue.

PR #255, opened 2025-04-24
Generalize ZeroPadding

LindaGuiga:
This PR aims at generalizing the zero padding. 
Arguments:
- num_extra_variables: number of additional variables.
- start_index: variable index (within the set of all variables) where the padding starts
- nonzero_index: the index (between 0 and (1 << num_extra variables) - 1) of the block (of size 1 << start_index) where the values are those of the original column.

For public access, I added the following:
`pub fn add_zero_pad_upcast<FSub, const VALUES_PER_ROW: usize, const NEW_VALUES_PER_ROW: usize>(
		&mut self,
		name: impl ToString,
		col: Col<FSub, VALUES_PER_ROW>,
	) -> Col<FSub, NEW_VALUES_PER_ROW>`

This is a method that can be use to easily go, for example, from `Col<B1, 16>` to `Col<B1, 32>`.


crates/circuits/src/builder/constraint_system.rs
```diff
@@ -415,11 +415,13 @@ impl<'arena> ConstraintSystemBuilder<'arena> {
 		name: impl ToString,
 		id: OracleId,
 		n_vars: usize,
+		nonzero_index: usize,
```
djadjka, line 418: could you please add rust doc for nonzero_index and start_index

crates/math/src/fold.rs
```diff
@@ -25,6 +25,59 @@ use stackalloc::helpers::slice_assume_init_mut;
...
+	P: PackedField,
+	PE: PackedField<Scalar: ExtensionField<P::Scalar>>,
+{
+	if start_index > log_evals_size {
+		bail!(Error::IncorrectStartIndex {
+			expected: log_evals_size
+		});
+	}
+
+	let extra_vals = log_new_vals - log_evals_size;
```
djadjka, line 46: maybe better extra_vars, or log_extra_vals

crates/math/src/fold.rs
```diff
@@ -25,6 +25,59 @@ use stackalloc::helpers::slice_assume_init_mut;
...
+	if nonzero_index > 1 << extra_vals {
```
djadjka, line 47: I'm not sure that i'm understand this constraint 
I think [1,2,3,0,0,4,5,6,0] is correct example, 
where extra_vals = 3 and nonzero_index (1 << 5) -1

maybe you mean nonzero_index > 1 << (log_new_vals - (log_evals_size - start_index))

djadjka (commented):
Nice work! 
I left a few comments — the most important one is to try to avoid using get/set when possible. But maybe we can do it as separate ticket 
cc @jimpo

crates/math/src/fold.rs
LindaGuiga: Sorry, I'm not sure I understand what you mean: if we have `m` extra variables `Y = (Y_0, ..., Y_{m-1})` then the nonzero_index indicates a value for `Y` that we want to select, so `nonzero_index` has to be between `0` and `2^m - 1`. Here, `extra_vals` is `m` (granted, I should change the name to `extra_vars`, sorry)

djadjka: Oh, nonzero_index counts from 1 << start_index not from 0
Sorry for confusing you

LindaGuiga: Oh so we want `nonzero_index` to be between `1 << start_index` and `1 << start_index + 1 << m - 1`?
But the argument for SelectRow would still be `nonzero_index - 1 << start_index` then, right? (since we're applying it to the extra variables directly)

jimpo: No, nonzero_index should start at 0

crates/math/src/fold.rs
```diff
@@ -25,6 +25,59 @@ use stackalloc::helpers::slice_assume_init_mut;
 
 use crate::Error;
 
+pub fn zero_pad<P, PE>(
```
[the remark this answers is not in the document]
LindaGuiga: I'm sorry, I'm not sure I understand: do you mean the cases where we can directly set slices instead of going one scalar element at a time?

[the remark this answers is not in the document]
djadjka: yes

crates/core/src/oracle/multilinear.rs
```diff
@@ -673,6 +686,40 @@ impl<F: TowerField> Projected<F> {
 	}
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Getters, CopyGetters, SerializeBytes, DeserializeBytes)]
+pub struct ZeroPadded {
+	#[get_copy = "pub"]
+	id: OracleId,
+	#[get_copy = "pub"]
+	new_n_vars: usize,
```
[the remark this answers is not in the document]
djadjka: Sorry, It was my suggestion, 
IMO `new_n_vars` more intuitive

jimpo (approved):

PR #257, opened 2025-04-25
[gadgets] Signed Multiplication Gadgets

Anex007:
- This MR adds `MulSU32`, `MulSS32`, `TwosComplement`, `Incr`, `SignConverter` gadgets.
- Added better testing which can possibly be reused in other gadgets.

### DIFF
--- a/crates/m3/src/gadgets/mul.rs
+++ b/crates/m3/src/gadgets/mul.rs
@@ -466,7 +466,7 @@ impl MulSS32 {
 		let new_prod_high_bits = array::from_fn(|bit| {
 			table.add_computed(
 				format!("new_prod_high[{bit}]"),
-				prod_high_bits[bit] * (x_is_negative + B1::ONE) + y_sub.zout[bit] * x_is_negative,
+				prod_high_bits[bit] + x_is_negative * (prod_high_bits[bit] + y_sub.zout[bit]),
 			)
 		});
 
@@ -484,8 +484,8 @@ impl MulSS32 {
 		let out_high_bits: [_; 32] = array::from_fn(|bit| {
 			table.add_computed(
 				format!("out_high[{bit}]"),
-				new_prod_high_bits[bit] * (y_is_negative + B1::ONE)
-					+ x_sub.zout[bit] * y_is_negative,
+				new_prod_high_bits[bit]
+					+ y_is_negative * (new_prod_high_bits[bit] + x_sub.zout[bit]),
 			)
 		});
 
@@ -634,7 +634,7 @@ impl MulSU32 {
 		let out_high_bits = array::from_fn(|bit| {
 			table.add_computed(
 				format!("out_high[{bit}]"),
-				prod_high_bits[bit] * (x_is_negative + B1::ONE) + y_sub.zout[bit] * x_is_negative,
+				prod_high_bits[bit] + x_is_negative * (prod_high_bits[bit] + y_sub.zout[bit]),
 			)
 		});
 
--- a/crates/m3/src/gadgets/mul.rs
+++ b/crates/m3/src/gadgets/mul.rs
@@ -5,7 +5,7 @@ use std::{array, marker::PhantomData};
 use anyhow::Result;
 use binius_field::{
 	ext_basis, packed::set_packed_slice, BinaryField, ExtensionField, Field, PackedExtension,
-	PackedField, PackedFieldIndexable, PackedSubfield, TowerField,
+	PackedField, PackedSubfield, TowerField,
 };
 use itertools::izip;
 
@@ -400,7 +400,9 @@ impl MulUU64 {
 	}
 }
 
-fn pack_fp<FP: TowerField, const BIT_LENGTH: usize>(bits: [Col<B1>; BIT_LENGTH]) -> Expr<FP, 1> {
+pub(crate) fn pack_fp<FP: TowerField, const BIT_LENGTH: usize>(
+	bits: [Col<B1>; BIT_LENGTH],
+) -> Expr<FP, 1> {
 	assert_eq!(BIT_LENGTH, 1 << FP::TOWER_LEVEL);
 	let basis: [_; BIT_LENGTH] = array::from_fn(ext_basis::<FP, B1>);
 	bits.into_iter()
@@ -420,8 +422,9 @@ pub struct MulSS32 {
 	out_signed_value: SignConverter<u64, 64>,
 	x_abs_bits: [Col<B1>; 32],
 	y_abs_bits: [Col<B1>; 32],
-	out_bits: [Col<B1>; 64],
 
+	// Outputs
+	pub out_bits: [Col<B1>; 64],
 	pub x_in: Col<B32>,
 	pub y_in: Col<B32>,
 	pub out_high: Col<B32>,
@@ -433,6 +436,14 @@ impl MulSS32 {
 		let x_in_bits = table.add_committed_multiple("x_in_bits");
 		let y_in_bits = table.add_committed_multiple("y_in_bits");
 
+		Self::with_input(table, x_in_bits, y_in_bits)
+	}
+
+	pub fn with_input(
+		table: &mut TableBuilder,
+		x_in_bits: [Col<B1>; 32],
+		y_in_bits: [Col<B1>; 32],
+	) -> Self {
 		let x_in = table.add_computed("SS::x_in", pack_fp(x_in_bits));
 		let y_in = table.add_computed("SS::y_in", pack_fp(y_in_bits));
 
@@ -489,7 +500,7 @@ impl MulSS32 {
 		y_vals: impl IntoIterator<Item = B32>,
 	) -> Result<()>
 	where
-		P: PackedFieldIndexable<Scalar = B128> + PackedExtension<B1> + PackedExtension<B32>,
+		P: PackedField<Scalar = B128> + PackedExtension<B1> + PackedExtension<B32>,
 	{
 		let mut inner_mul_x = Vec::new();
 		let mut inner_mul_y = Vec::new();
@@ -634,7 +645,7 @@ impl MulSU32 {
 		y_vals: impl IntoIterator<Item = B32>,
 	) -> Result<()>
 	where
-		P: PackedFieldIndexable<Scalar = B128> + PackedExtension<B1> + PackedExtension<B32>,
+		P: PackedField<Scalar = B128> + PackedExtension<B1> + PackedExtension<B32>,
 	{
 		let mut mul_inner_x = Vec::new();
 		let mut mul_inner_y = Vec::new();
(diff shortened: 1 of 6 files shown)


jimpo (commented):
Skimmed it. The approach with 2s complement conversion looks fine to me.

@pepyakin Can you please review the constraint system code?

crates/m3/src/gadgets/u32/add.rs
```diff
@@ -134,6 +142,112 @@ impl U32Add {
...
+		let cout: [_; BIT_LENGTH] = array::from_fn(|i| {
+			if i != BIT_LENGTH - 1 {
+				cin[i + 1]
+			} else {
+				final_carry_out
+			}
+		});
+
+		for (i, &xin) in input.iter().enumerate() {
+			table.assert_zero(format!("sum_{i}"), xin + cin[i] - zout[i]);
```
pepyakin, line 199: Idk much about the conventions about feels like the indexing notation would be more consistent here.

```suggestion
			table.assert_zero(format!("sum[{i}]"), xin + cin[i] - zout[i]);
```

crates/m3/src/gadgets/u32/add.rs
```diff
@@ -134,6 +142,112 @@ impl U32Add {
...
+		for (final_carry, carry) in izip!(final_carry_out.iter_mut(), carry_in.iter()) {
+			*final_carry = *carry;
+		}
+
+		Ok(())
+	}
+}
+
+pub type U32Incr = Incr<u32, 32>;
+pub type U64Incr = Incr<u64, 64>;
```
pepyakin, line 249: would be good to have tests.

crates/m3/src/gadgets/u32/add.rs
```diff
@@ -134,6 +142,112 @@ impl U32Add {
...
+#[derive(Debug)]
+pub struct Incr<UPrimitive: UnsignedAddPrimitives, const BIT_LENGTH: usize> {
+	cin: [Col<B1>; BIT_LENGTH],
+
+	/// Input column as bits.
+	pub input: [Col<B1>; BIT_LENGTH],
+	/// Output column as bits.
+	pub zout: [Col<B1>; BIT_LENGTH],
+	/// Output carry bit if there is any.
+	pub final_carry_out: Col<B1>,
```
pepyakin, line 173: It's a bit unfortunate that this is unconditional, unlike say in `U32Add`.

pepyakin (commented):
did a quick glance. Will do another look tomorrow.

crates/field/src/arch/aarch64/m128.rs
```diff
@@ -248,7 +248,7 @@ impl std::fmt::Display for M128 {
 
 impl std::fmt::Debug for M128 {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		write!(f, "M128({})", self)
+		write!(f, "M128({self})")
```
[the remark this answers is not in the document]
Anex007: clippy was throwing me errors, this was missed in #252 because of conditional compilation by platforms

crates/m3/src/gadgets/mul.rs
```diff
@@ -407,3 +411,408 @@ fn pack_fp<FP: TowerField, const BIT_LENGTH: usize>(bits: [Col<B1>; BIT_LENGTH])
...
+	///
+	/// ## Example
+	/// - If the conditional is zero, the output will be the input bits.
+	/// - If the conditional is one, the output will be the two's complement of input bits.
+	///
+	pub fn new(
+		table: &mut TableBuilder,
+		name: &str,
+		in_bits: [Col<B1>; BIT_LENGTH],
+		conditional: Expr<B1, 1>,
```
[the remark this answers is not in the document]
Anex007: In #258 we use this gadget to perform subtraction of the form `|r| - |q|` so the conditional bit in that case would be sign bit plus one.

crates/m3/src/gadgets/mul.rs
```diff
@@ -78,7 +80,7 @@ impl UnsignedMulPrimitives for u64 {
 
 	fn shifted_generator() -> B128 {
 		let mut g = Self::generator();
-		for _ in 0..Self::BIT_LENGTH {
+		for _ in 0..64 {
```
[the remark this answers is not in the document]
Anex007: After adding the trait `UnsignedAddPrimitives` the compiler wanted to be explicit on which associated value I was referencing, it was less convoluted to specify as a fixed value.

crates/m3/src/gadgets/mul.rs
```diff
@@ -398,12 +403,407 @@ impl MulUU64 {
...
+			let mut out_bits = array_util::try_map(self.out_bits, |bit| index.get_mut(bit))?;
+			let mut x_abs_bits = array_util::try_map(self.x_abs_bits, |bit| index.get_mut(bit))?;
+
+			let mut x_in = index.get_mut(self.x_in)?;
+			let mut y_in = index.get_mut(self.y_in)?;
+			let mut out_low = index.get_mut(self.out_low)?;
+			let mut out_high = index.get_mut(self.out_high)?;
+
+			for (i, (x, y)) in x_vals.into_iter().zip(y_vals.into_iter()).enumerate() {
+				let x_i32 = i32::from_le_bytes(x.val().to_le_bytes());
```
pepyakin, line 655: I am a little bit confused. 

Here, we are reinterpretting the bytes of an unsigned 32-bit integer `x.val()` in little endian and then creating a signed 32-bit integer from those LE bytes.

While correct (even if the machine is not an LE I believe), this is probably better be performed as a simple cast, as is done for `MulSS32`. Or am I missing something?

crates/m3/src/gadgets/mul.rs
```diff
@@ -398,12 +403,407 @@ impl MulUU64 {
...
+
+		Ok(())
+	}
+}
+
+/// Simple struct to convert to and from Two's complement representation based on bits. See [`SignConverter::new`]
+///
+/// NOTE: *We do not handle witness generation for the `converted_bits` and should be handled by caller*
+#[derive(Debug)]
+pub struct SignConverter<UPrimitive: UnsignedAddPrimitives, const BIT_LENGTH: usize> {
```
pepyakin, line 709: is this just `Abs`? If so, wouldn't it be better, at least for discoverability sake, to name it this way?

crates/m3/src/gadgets/mul.rs
```diff
@@ -398,12 +403,407 @@ impl MulUU64 {
...
+		self.y_abs_value.populate(index)?;
+		self.mul_inner.populate(index, inner_mul_x, inner_mul_y)?;
+		self.out_signed_value.populate(index)?;
+
+		Ok(())
+	}
+}
+
+#[derive(Debug)]
+pub struct MulSU32 {
```
pepyakin, line 571: I think this module could use some tests. Bonus points for focusing on interesting cases in addition ot the property/RNG tests.

crates/m3/src/gadgets/mul.rs
Anex007: You're right this is something I missed when changing all my previous code to to use simple casting, I was testing code with byte conversions to hunt down a bug

crates/m3/src/gadgets/mul.rs
```diff
@@ -407,3 +411,408 @@ fn pack_fp<FP: TowerField, const BIT_LENGTH: usize>(bits: [Col<B1>; BIT_LENGTH])
...
+	pub x_in: [Col<B1>; BIT_LENGTH],
+	pub result_bits: [Col<B1>; BIT_LENGTH],
+}
+
+impl<UPrimitive: UnsignedAddPrimitives, const BIT_LENGTH: usize>
+	TwosComplement<UPrimitive, BIT_LENGTH>
+{
+	pub fn new(table: &mut TableBuilder, x_in: [Col<B1>; BIT_LENGTH]) -> Self {
+		let inverted = array::from_fn(|i| {
+			table.add_computed(format!("TwosComplement::inverted[{i}]"), x_in[i] + B1::ONE)
```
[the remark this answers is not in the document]
Anex007: `TableBuilder::with_namespace` overwrites the namespace, and there is no way to "pop" the namespace without storing a list of namespaces. So if I were to call another gadget after calling this `TwosComplement` gadget the namespace will be appended to it rather than overwriting it. We should modify name-spacing to what we have in `ConstraintSystemBuilder::push_namespace`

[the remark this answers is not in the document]
pepyakin: Oh, that would've been confusing! I checked however and luckily it is not the case, the test below passed:

```
	#[test]
	fn namespace_nesting() {
		let mut table = Table::<B128>::new(0, "table");
		let mut tb = TableBuilder::new(&mut table);
		let mut tb_ns_1 = tb.with_namespace("ns1");
		let tb_ns_2 = tb_ns_1.with_namespace("ns2");
		assert_eq!(tb_ns_2.namespaced_name("column"), "ns1::ns2::column");
	}
```

IOW, there is no need to push/pop namespaces as this is naturally achieved by using namespaced `TableBuilder`s. And, again AFAICT, the convention is is that it's the caller responsibility to assign the namespace.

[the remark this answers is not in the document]
Anex007: Ah I see what you mean, will fix that!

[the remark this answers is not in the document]
pepyakin: FWIW, added this to clarify https://github.com/IrreducibleOSS/binius/pull/285

crates/m3/src/gadgets/mul.rs
Anex007: Added some explicit tests let me know if you want me to add more

crates/m3/src/gadgets/mul.rs
```diff
@@ -398,12 +407,416 @@ impl MulUU64 {
...
+			U32SubFlags {
+				commit_zout: true,
+				..Default::default()
+			},
+		);
+
+		let new_prod_high_bits = array::from_fn(|bit| {
+			table.add_computed(
+				format!("new_prod_high[{bit}]"),
+				prod_high_bits[bit] * (x_is_negative + B1::ONE) + y_sub.zout[bit] * x_is_negative,
```
jimpo, line 469: You can factor this expression to do just one multiplication. Same with the other conditional.

Perhaps it'd be worth adding a utility function for building a conditionally selected `Expr`.

crates/m3/src/gadgets/mul.rs
```diff
@@ -398,12 +403,407 @@ impl MulUU64 {
...
+		self.mul_inner.populate(index, mul_inner_x, mul_inner_y)?;
+		self.out_signed_value.populate(index)?;
+
+		Ok(())
+	}
+}
+
+/// Simple struct to convert to and from Two's complement representation based on bits. See [`SignConverter::new`]
+///
+/// NOTE: *We do not handle witness generation for the `converted_bits` and should be handled by caller*
```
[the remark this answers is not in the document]
jimpo: Sounds good

jimpo (commented):
Nice, thanks for updating this! Looks good, just one comment.

crates/m3/src/gadgets/mul.rs
Anex007: That's a good point

jimpo (approved):

PR #258, opened 2025-04-25
[gadgets] Division gadgets for M3

Anex007:
- This MR adds `DivUU32`, `DivSS32` and non packed addition variant for u32 and u64 addition called `WideAdd` 
- This MR depends upon #257 to be merged first

### DIFF
--- a/crates/m3/src/gadgets/div.rs
+++ b/crates/m3/src/gadgets/div.rs
@@ -3,6 +3,7 @@
 use std::array;
 
 use binius_field::{packed::set_packed_slice, Field, PackedExtension, PackedField};
+use binius_maybe_rayon::prelude::IntoParallelIterator;
 use itertools::izip;
 
 use crate::{
@@ -35,7 +36,6 @@ pub struct DivUU32 {
 }
 
 impl DivUU32 {
-	#[allow(clippy::needless_range_loop)]
 	pub fn new(table: &mut TableBuilder) -> Self {
 		let zero = table.add_constant("zero", [B1::ZERO]);
 		let p_in_bits = table.add_committed_multiple("p_in_bits");
@@ -80,9 +80,10 @@ impl DivUU32 {
 			},
 		);
 
+		#[allow(clippy::needless_range_loop)]
 		for bit in 0..64 {
 			table.assert_zero(
-				format!("division_satisfied_{bit}"),
+				format!("division_satisfied[{bit}]"),
 				zero_extend_p[bit] - sum.z_out[bit],
 			);
 		}
@@ -215,7 +216,6 @@ pub struct DivSS32 {
 }
 
 impl DivSS32 {
-	#[allow(clippy::needless_range_loop)]
 	pub fn new(table: &mut TableBuilder) -> Self {
 		let p_in_bits = table.add_committed_multiple("p_in_bits");
 		let q_in_bits = table.add_committed_multiple("q_in_bits");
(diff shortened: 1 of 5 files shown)


crates/m3/src/gadgets/div.rs
```diff
@@ -0,0 +1,414 @@
...
+	out_rem_bits: [Col<B1>; 32],
+
+	pub p_in: Col<B32>,
+	pub q_in: Col<B32>,
+	pub out_div: Col<B32>,
+	pub out_rem: Col<B32>,
+}
+
+impl DivUU32 {
+	#[allow(clippy::needless_range_loop)]
```
pepyakin, line 38: I would suggest moving this to the loop itself:

```rust
#[allow(clippy::needless_range_loop)]
for bit in 0..64 {
  // .. 
}
```

crates/m3/src/gadgets/div.rs
```diff
@@ -0,0 +1,414 @@
...
+			U32AddFlags {
+				commit_zout: true,
+				expose_final_carry: true,
+				..Default::default()
+			},
+		);
+
+		for bit in 0..64 {
+			table.assert_zero(
+				format!("division_satisfied_{bit}"),
```
pepyakin, line 85: Just to be consistent with index notation we use elsewhere.

```suggestion
				format!("division_satisfied[{bit}]"),
```

crates/m3/src/gadgets/u32/add.rs
```diff
@@ -142,6 +142,120 @@ impl U32Add {
...
+					.carry_in_bit
+					.map(|carry_col| table.add_selected("cin[0]", carry_col, 0))
+					.unwrap_or_else(|| table.add_committed("cin[0]"))
+			}
+		});
+		let zout = if flags.commit_zout {
+			let zout = table.add_committed_multiple("z_out");
+			for bit in 0..BIT_LENGTH {
+				table.assert_zero(
+					format!("sum_{bit}"),
```
pepyakin, line 182:
```suggestion
					format!("sum[{bit}]"),
```

crates/m3/src/gadgets/div.rs
```diff
@@ -0,0 +1,414 @@
...
+	pub fn populate_with_inputs<P>(
+		&self,
+		index: &mut TableWitnessSegment<P>,
+		p_vals: impl IntoIterator<Item = B32>,
+		q_vals: impl IntoIterator<Item = B32> + Clone,
+	) -> anyhow::Result<()>
+	where
+		P: PackedField<Scalar = B128> + PackedExtension<B1> + PackedExtension<B32>,
+	{
+		let mut inner_div = Vec::new();
```
pepyakin, line 328: Would be good to leave a note of what've gathering here as a comment.

crates/m3/src/gadgets/div.rs
```diff
@@ -0,0 +1,414 @@
...
+				};
+				set_packed_slice(&mut p_in, i, p);
+				set_packed_slice(&mut q_in, i, q);
+				let div = B32::new(div as u32);
+				let rem = B32::new(rem as u32);
+				set_packed_slice(&mut out_div, i, div);
+				set_packed_slice(&mut out_rem, i, rem);
+				inner_div.push(div);
+
+				for bit_idx in 0..32 {
```
pepyakin, line 368: Is it me or this loop looks terribly inefficient? Is there a way around it? 

Not blocking

pepyakin (approved):
This looks broadly ok.

crates/m3/src/gadgets/div.rs
Anex007: Yeah the double inner loop is not desirable, but I dont see any other efficient way to populate the witness given that we have to take in a iterator of values to divide, I can make it make a parallel iterator though.

jimpo (approved):

PR #266, opened 2025-04-28
[compute] Extend with functionality for multilinear evaluation

iradicek:
Extends `ComputeLayer` with test suite (generalised over compute layer) for some multilinear evaluation and needed functionality in the layer trait itself.


crates/compute/src/tower.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2024-2025 Irreducible Inc.
+
+//! Traits for working with field towers.
```
jimpo, line 3: I think we should probably just move `tower.rs` from `binius_core` to `binius_field` now. @GraDKh agree, disagree?

crates/compute/src/cpu/layer.rs
```diff
@@ -50,4 +62,276 @@ impl<F: TowerField> ComputeLayer<F> for CpuLayer<F> {
...
+			.take(1 << (n_vars - <BinaryField128b as ExtensionField<BinaryField32b>>::LOG_DEGREE))
+			.collect::<Vec<_>>();
+		let coordinates = repeat_with(|| <BinaryField128b as Field>::random(&mut rng))
+			.take(n_vars)
+			.collect::<Vec<_>>();
+
+		let mut eq_ind_buffer = zeroed_vec(1 << n_vars);
+
+		#[allow(clippy::too_many_arguments)]
+		fn compute_multiple_multilinear_evaluations<F, C: ComputeLayer<F>>(
```
jimpo, line 267: Slight preference to move these nested functions out to the `tests` module level.

crates/compute/src/cpu/layer.rs
```diff
@@ -50,4 +62,276 @@ impl<F: TowerField> ComputeLayer<F> for CpuLayer<F> {
...
+	use rand::{prelude::StdRng, SeedableRng};
+
+	use super::*;
+	use crate::{memory::ComputeMemory, tower::CanonicalTowerFamily};
+
+	#[test]
+	fn test_exec_single_tensor_expand() {
+		let n_vars = 8;
+		let mut rng = StdRng::seed_from_u64(0);
+		let coordinates = repeat_with(|| <BinaryField128b as Field>::random(&mut rng))
```
jimpo, line 172: More of this code (ideally almost everything except creating the `hal`) should go into the shared test. Use `copy_h2d` to copy to a device buffer. Same with the other tests. See the memory tests in `src/layer.rs` for example.

crates/compute/src/cpu/layer.rs
```diff
@@ -165,40 +165,60 @@ mod tests {
...
 		tensor_prod_eq_ind(2, &mut buffer_clone, &coordinates[2..]).unwrap();
 		assert_eq!(buffer, buffer_clone);
 	}
 
+	#[test]
+	fn test_exec_single_tensor_expand() {
+		type F = BinaryField128b;
+		let n_vars = 8;
+		let compute = <CpuLayer<CanonicalTowerFamily>>::default();
+		let mut device_memory = vec![F::ONE; 1 << n_vars];
```
jimpo, line 218: Probably better to initialize with 0s than 1s? I suppose it's arbitrary, but that seems more natural.

jimpo (commented):
The generic tensor_expand test looks good

crates/compute/src/tower.rs
GraDKh: Totally agree with that

crates/compute/src/cpu/layer.rs
iradicek: I intentionally used 1s because it seemed like it would be more likely to cause test failure than 0s. For example, one of the steps is to set some memory to 0, if you initialize with 0s and miss that step the test will still pass, while with 1s it will fail. However, this as well might be overthinking and not needed. Will revert to 0s...

jadnohra: I agree that not using 0s is often a better way to catch bugs. I don't know if that applies here though.

jimpo:
Nice job! Please rebase onto main to use `binius_field::tower`, then this is good to go.

crates/compute/src/tower.rs
iradicek: Removed and using field tower now.

iradicek:
> Nice job! Please rebase onto main to use `binius_field::tower`, then this is good to go.

Thanks! Rebased and using field tower

jimpo (approved):

PR #269, opened 2025-04-28
[m3] Add structured column type

jimpo:
This adds dynamically-sized structured columns. In a future change, I will add fixed-size structured columns as well.

The first structured column type is incremented value columns. These will be used for zCray address spaces.

### DIFF
--- a/crates/m3/Cargo.toml
+++ b/crates/m3/Cargo.toml
@@ -21,6 +21,7 @@ either.workspace = true
 getset.workspace = true
 itertools.workspace = true
 thiserror.workspace = true
+log = "0.4.27"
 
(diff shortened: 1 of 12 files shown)


crates/m3/Cargo.toml
```diff
@@ -21,6 +21,7 @@ either.workspace = true
 getset.workspace = true
 itertools.workspace = true
 thiserror.workspace = true
+log = "0.4.27"
```
djadjka, line 24: Should it be removed?

djadjka (commented):
LGTM, just question about `log` crate in m3 Cargo.toml

PR #277, opened 2025-04-29
[evalcheck] Write duplicate claim advice directly to transcript

jimpo:
This removes most of the serialization/deserialization & proof struct pattern matching logic from evalcheck. This simplifies the code a lot. Documentation needs updating.

### DIFF
--- a/crates/core/src/protocols/evalcheck/prove.rs
+++ b/crates/core/src/protocols/evalcheck/prove.rs
@@ -7,12 +7,13 @@ use binius_hal::ComputationBackend;
 use binius_math::MultilinearExtension;
 use binius_maybe_rayon::prelude::*;
 use getset::{Getters, MutGetters};
-use itertools::izip;
+use itertools::{chain, izip};
 use tracing::instrument;
 
 use super::{
 	error::Error,
 	evalcheck::{EvalcheckMultilinearClaim, EvalcheckProof},
+	serialize_evalcheck_proof,
 	subclaims::{
 		add_composite_sumcheck_to_constraints, calculate_projected_mles, composite_sumcheck_meta,
 		fill_eq_witness_for_composites, MemoizedData, ProjectedBivariateMeta,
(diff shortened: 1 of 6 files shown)


crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -340,6 +342,16 @@ where
...
 							* coeff.invert().expect("not zero");
+
+						if self
+							.evals_memoization
+							.get(suboracle_id, &eval_point)
+							.is_none()
+						{
+							self.evals_memoization
+								.insert(suboracle_id, eval_point.clone(), eval);
+						}
```
djadjka, line 353:
```suggestion
let eval = if let Some(eval) = self.evals_memoization.get(suboracle_id, &eval_point) {
    *eval
} else {
    let eval = (eval - linear_combination.offset())
        * coeff.invert().expect("not zero");
    self.evals_memoization.insert(suboracle_id, eval_point.clone(), eval);
    eval
};
```

crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -372,21 +384,37 @@ where
...
+	fn prove_multilinear<Challenger_: Challenger>(
 		&mut self,
 		evalcheck_claim: EvalcheckMultilinearClaim<F>,
-	) -> Result<EvalcheckProof<F>, Error> {
-		let EvalcheckMultilinearClaim {
-			id,
-			eval_point,
-			eval,
-		} = evalcheck_claim.clone();
+		transcript: &mut ProverTranscript<Challenger_>,
```
djadjka, line 390: I think we can add transcript to the EvalcheckProver and call `self.transcript`  

and it's allow to do something like this 

```
pub fn add_duplicate_claim_to_proof(&mut self, claim_id: usize) {
	serialize_evalcheck_proof(
		&mut self.transcript.message(),
		&EvalcheckProof::DuplicateClaim(claim_id),
	);
}

pub fn add_new_claim_to_proof(&mut self) {
	serialize_evalcheck_proof(&mut self.transcript.message(), &EvalcheckProof::NewClaim);
}
```

djadjka (commented):
Okay, I was wrong — there is still room for improvement in evalcheck :smiling_face_with_tear:

crates/core/src/protocols/evalcheck/prove.rs
jimpo: There's already too many fields on `EvalcheckProver` -- I'd prefer to have it as an explicit argument. It also matches how transcript is used elsewhere in the code.

djadjka (approved):

Issue #265, opened 2025-04-28
Optimize M3 ConstraintSet compilation


jimpo:
Our new Keccak M3 circuit identified a performance bug with the [compilation](https://github.com/IrreducibleOSS/binius/blob/e81c4bcd41bb8df6cf01037f7b669f84c603ad76/crates/m3/src/builder/constraint_system.rs#L167) of M3 ConstraintSystems to binius_core CompiledConstraintSystems. Specifically, [the way the compile method constructs ConstraintSets](https://github.com/IrreducibleOSS/binius/blob/e81c4bcd41bb8df6cf01037f7b669f84c603ad76/crates/m3/src/builder/constraint_system.rs#L323) is overly conservative and includes multilinear oracles in the ConstraintSet that have no zero constraints.

In the Keccak circuit, this results in ~150 unconstrained columns getting included in zerocheck, compared with the old version. The extra folding computation performed is the performance regression.

The easiest/best solution I can think of is to filter out the oracles that are not referenced by any constraints, and accordingly [remap the variable indices](https://github.com/IrreducibleOSS/binius/blob/2affbc12e2880d3c32c69126eb2fee178dc70ffc/crates/math/src/arith_expr.rs#L122) in the ArithExpr representing the constraints.

PR #278, opened 2025-04-30
[m3] use only referenced oracle ids for zero check

pepyakin:
Instead of unconditionally adding the whole list of per partition oracle ids
into the constraint set, we check which ones are actually getting used.

See [CRY-296]

Fixes https://github.com/IrreducibleOSS/binius/issues/265

[CRY-296]: https://linear.app/irreducible/issue/CRY-297/optimize-m3-constraintset-compilation


jimpo:
Fixes https://github.com/IrreducibleOSS/binius/issues/265

jimpo (approved):

PR #282, opened 2025-04-30
[fri] Simplify the FRI folding code

jimpo:
This simplifies the FRI folding code in a few ways. The improvements are:

- Remove the requirement that fold_challenges be non-empty
- Remove the `fold_codeword` prover function and use the more general `fold_interleaved` in all cases
- Pass `log_len` instead of `round` parameter so that the code will work with NTT instances with larger domains

### DIFF
--- a/crates/core/src/protocols/fri/common.rs
+++ b/crates/core/src/protocols/fri/common.rs
@@ -1,15 +1,13 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::marker::PhantomData;
+use std::{iter, marker::PhantomData};
 
-use binius_field::{
-	packed::len_packed_slice, util::inner_product_unchecked, BinaryField, ExtensionField,
-	PackedField,
-};
+use binius_field::{util::inner_product_unchecked, BinaryField, ExtensionField, PackedField};
 use binius_math::extrapolate_line_scalar;
 use binius_ntt::AdditiveNTT;
 use binius_utils::bail;
 use getset::{CopyGetters, Getters};
+use itertools::Itertools;
 
 use crate::{
 	merkle_tree::MerkleTreeScheme, protocols::fri::Error,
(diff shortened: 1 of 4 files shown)


crates/core/src/protocols/fri/common.rs
```diff
@@ -38,66 +36,69 @@ where
...
-		} else {
-			// For the first round, we read values directly from the `values` slice.
-			(0..new_scratch_buffer_len).for_each(|index_offset| {
-				let values = (values[index_offset << 1], values[(index_offset << 1) + 1]);
-				scratch_buffer[index_offset] =
-					fold_pair(ntt, round, index_start + index_offset, values, r)
-			});
+		let ntt_round = ntt.log_domain_size() - log_len;
+		for index_offset in 0..1 << (log_size - 1) {
+			let pair = (values[index_offset << 1], values[(index_offset << 1) + 1]);
```
benediamond, line 87:
```suggestion
			let pair = (values[index_offset << 1], values[index_offset << 1 | 1]);
```
not sure if it matters? but in general bitwise OR prob more efficient than add?

crates/core/src/protocols/fri/common.rs
```diff
@@ -38,66 +36,69 @@ where
...
+			values[index_offset] = fold_pair(
+				ntt,
+				ntt_round,
+				(chunk_index << (log_size - 1)) + index_offset,
```
benediamond, line 91:
```suggestion
				chunk_index << log_size - 1 | index_offset,
```
similar point here

crates/core/src/protocols/fri/common.rs
jimpo: Not on CPUs. And our linter requires parenthesis here :-)

benediamond:
looks good; a few comments:
- your current implementation of [`fold_interleaved_chunk`](https://github.com/IrreducibleOSS/binius/blob/70ca1174b0def866dbba43773495275c06e52a3d/crates/core/src/protocols/fri/common.rs#L128-L137) appears to support a sort of general situation in which `log_batch_size != 0` _and_ `fold_challenges` is nonempty. that is, we will do _some_ interleaved folding, and then _some more_ noninterleaved folding. does this code path ever get used?
- you do your interleaved and noninterleaved folding differently: interleaved folding is done by dotting with a pre-expanded tensor; while noninterleaved is done in a "round after round" way.
- as I mentioned, I structured the code in a different way: folding interleaved and noninterleaved are done with just one method. both completely consume and use up a list of challenges. both are done in a "round after round" way. the only difference is whether we interleave each round with an inverse butterfly stage (we do this if and only if we are doing noninterleaved folding). this is controlled by a single boolean flag passed in.

i don't think there's anything wrong with what you're doing here; and also there are some subtleties with `P::LOG_WIDTH` which i unfortunately don't grok. but I think that my way would lead to less overall code? overall i'm fine with this though; just want to make sure you're aware of the other way

jimpo:
>     * your current implementation of [`fold_interleaved_chunk`](https://github.com/IrreducibleOSS/binius/blob/70ca1174b0def866dbba43773495275c06e52a3d/crates/core/src/protocols/fri/common.rs#L128-L137) appears to support a sort of general situation in which `log_batch_size != 0` _and_ `fold_challenges` is nonempty. that is, we will do _some_ interleaved folding, and then _some more_ noninterleaved folding. does this code path ever get used?

Sort of. It's a valid supported pattern that the FRI committer doesn't know about / think about fold arities when committing, then the prover chooses a larger fold arity than the log batch size. In the PCS itself though, the committer does know about the fold arities and chooses the batch size to equal the first fold arity.

>     * you do your interleaved and noninterleaved folding differently: interleaved folding is done by dotting with a pre-expanded tensor; while noninterleaved is done in a "round after round" way.
> 
>     * as I mentioned, I structured the code in a different way: folding interleaved and noninterleaved are done with just one method. both completely consume and use up a list of challenges. both are done in a "round after round" way. the only difference is whether we interleave each round with an inverse butterfly stage (we do this if and only if we are doing noninterleaved folding). this is controlled by a single boolean flag passed in.

We discussed offline. That's a good approach for the verifier & when we don't have packed fields. In the case of the prover folding using packed fields, my approach makes it somewhat easier to implement using packed multiplications. This is something we could change in the future.

benediamond (approved):

crates/core/src/protocols/fri/common.rs
```diff
@@ -139,36 +141,34 @@ where
...
+		values.len(),
+		1 << (fold_challenges.len() + log_batch_size).saturating_sub(P::LOG_WIDTH)
+	);
+	debug_assert_eq!(tensor.len(), 1 << log_batch_size.saturating_sub(P::LOG_WIDTH));
+	debug_assert!(scratch_buffer.len() >= 1 << fold_challenges.len());
+
+	let scratch_buffer = &mut scratch_buffer[..1 << fold_challenges.len()];
+
+	if log_batch_size == 0 {
+		iter::zip(&mut *scratch_buffer, P::iter_slice(values)).for_each(|(dst, val)| *dst = val);
```
benediamond, line 156: btw, i confirmed (by testing) that this case can now be removed. haven't tested the impact on performance

PR #290, opened 2025-05-01
[evalcheck] Handle composite MLEs separately from projected bivariate sumchecks

jimpo:
They are different and should be handled differently. Composite MLE evaluations are not bivariate, and they don't use projection. In the future, we shouldn't handle them as regular sumchecks, we should prove them as MLEchecks, which have a more efficient proving algorithm.


djadjka (approved):

jimpo:
They are different and should be handled differently. Composite MLE evaluations are not bivariate, and they don't use projection. In the future, we shouldn't handle them as regular sumchecks, we should prove them as MLEchecks, which have a more efficient proving algorithm.

PR #293, opened 2025-05-02
SYS-220: Define HAL operation for left_fold

BenjaminTrapani:
Implement for the CPU reference layer. Add a new type combining an FSlice with a subfield degree.

### DIFF
--- a/crates/compute/src/cpu/layer.rs
+++ b/crates/compute/src/cpu/layer.rs
@@ -7,13 +7,76 @@ use binius_field::{
 };
 
 use super::{memory::CpuMemory, tower_macro::each_tower_subfield};
-use crate::layer::{ComputeLayer, Error, FSlice, FSliceMut};
+use crate::{
+	layer::{ComputeLayer, Error, FSlice, FSliceMut},
+	memory::FSliceWithTowerLevel,
+};
 
 #[derive(Debug)]
 pub struct CpuExecutor;
 
 #[derive(Debug, Default)]
 pub struct CpuLayer<F: TowerFamily>(PhantomData<F>);
+
+impl<F: TowerFamily> CpuLayer<F> {
+	pub fn new() -> Self {
+		Self(PhantomData)
+	}
+}
+
+/// evals is treated as a matrix with `1 << log_query_size` rows and each column is dot-producted
+/// with the corresponding query element. The result is written to the `output` slice of packed values.
+fn compute_left_fold<EvalType: TowerField, T: TowerFamily>(
+	evals_as_b128: Vec<T::B128>,
+	log_evals_size: usize,
+	query: Vec<T::B128>,
+	log_query_size: usize,
+	out: FSliceMut<'_, T::B128, CpuLayer<T>>,
+) -> Result<(), Error>
+where
+	<T as TowerFamily>::B128: From<EvalType> + ExtensionField<EvalType>,
+{
+	let evals = evals_as_b128
+		.iter()
+		.flat_map(<T::B128 as ExtensionField<EvalType>>::iter_bases)
+		.collect::<Vec<_>>();
+	let num_rows = 1 << log_query_size;
+	let num_cols = 1 << (log_evals_size - log_query_size);
+
+	if evals.len() != num_rows * num_cols {
+		return Err(Error::InputValidation(format!(
+			"evals has {} elements, expected {}",
+			evals.len(),
+			num_rows * num_cols
+		)));
+	}
+
+	if query.len() != num_rows {
+		return Err(Error::InputValidation(format!(
+			"query has {} elements, expected {}",
+			query.len(),
+			num_rows
+		)));
+	}
+
+	if out.len() != num_cols {
+		return Err(Error::InputValidation(format!(
+			"output has {} elements, expected {}",
+			out.len(),
+			num_cols
+		)));
+	}
+
+	for i in 0..num_cols {
+		let mut acc = T::B128::ZERO;
+		for j in 0..num_rows {
+			acc += T::B128::from(evals[j * num_cols + i]) * query[j];
+		}
+		out[i] = acc;
+	}
+
+	Ok(())
+}
 
 impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
 	type Exec = CpuExecutor;
@@ -102,6 +165,28 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
 		Ok(result)
 	}
 
+	fn fold_left<'a>(
+		&'a self,
+		_exec: &'a mut Self::Exec,
+		evals: FSliceWithTowerLevel<'_, T::B128, <Self as ComputeLayer<T::B128>>::DevMem>,
+		log_evals_size: usize,
+		query: FSlice<'_, T::B128, Self>,
+		log_query_size: usize,
+		out: &mut FSliceMut<'_, T::B128, Self>,
+	) -> Result<(), Error> {
+		each_tower_subfield!(
+			evals.tower_level,
+			T,
+			compute_left_fold::<_, T>(
+				evals.slice.to_vec(),
+				log_evals_size,
+				query.to_vec(),
+				log_query_size,
+				out
+			)
+		)
+	}
+
 	fn tensor_expand(
 		&self,
 		_exec: &mut Self::Exec,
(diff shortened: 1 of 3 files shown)


crates/compute/src/memory.rs
```diff
@@ -72,3 +72,25 @@ pub trait ComputeMemory<F> {
 		Self::split_at_mut(borrowed, mid)
 	}
 }
+
+pub struct FSliceWithDegree<'a, F, Mem: ComputeMemory<F>> {
+	pub slice: Mem::FSlice<'a>,
+	pub degree: u8,
```
BenjaminTrapani, line 78: Per discussion with @benediamond, this should either be named `log_degree` or `tower_level`. `tower_level` seems clearer given the corresponding constant associated with `TowerField`.

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,77 @@ use binius_field::{
...
+	memory::FSliceWithTowerLevel,
+};
 
 #[derive(Debug)]
 pub struct CpuExecutor;
 
 #[derive(Debug, Default)]
 pub struct CpuLayer<F: TowerFamily>(PhantomData<F>);
 
+impl<F: TowerFamily> CpuLayer<F> {
```
nikurt, line 21: Not needed, because the user can do `CpuLayer::default()` to the exact same effect.

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,77 @@ use binius_field::{
...
+	pub fn new() -> Self {
+		Self(PhantomData)
+	}
+}
+
+/// evals is treated as a matrix with `1 << log_query_size` rows and each column is dot-producted
+/// with the corresponding query element. The result is written to the `output` slice of packed values.
+fn compute_left_fold<EvalType: TowerField, T: TowerFamily>(
+	evals_as_b128: Vec<T::B128>,
+	log_evals_size: usize,
```
nikurt, line 31: Please add a comment, or indicate otherwise, that evals can in fact be anything from 1b to 128b. Otherwise, the function signature looks like evals are exactly 128b elements.

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,77 @@ use binius_field::{
...
+	query: Vec<T::B128>,
+	log_query_size: usize,
+	out: FSliceMut<'_, T::B128, CpuLayer<T>>,
+) -> Result<(), Error>
+where
+	<T as TowerFamily>::B128: From<EvalType> + ExtensionField<EvalType>,
```
nikurt, line 37: I think you don't need `From`, because you can use `T::B128::cast_base()`.

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,77 @@ use binius_field::{
...
+{
+	let evals = evals_as_b128
+		.iter()
+		.flat_map(<T::B128 as ExtensionField<EvalType>>::iter_bases)
+		.collect::<Vec<_>>();
```
nikurt, line 42: In case of 1b eval elements, that would collect a `Vec` of `BinaryField1b` elements, which are definitely not `Pod`.

crates/compute/src/cpu/layer.rs
```diff
@@ -102,6 +165,28 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
...
+	fn fold_left<'a>(
+		&'a self,
+		_exec: &'a mut Self::Exec,
+		evals: FSliceWithTowerLevel<'_, T::B128, <Self as ComputeLayer<T::B128>>::DevMem>,
+		log_evals_size: usize,
+		query: FSlice<'_, T::B128, Self>,
+		log_query_size: usize,
+		out: &mut FSliceMut<'_, T::B128, Self>,
+	) -> Result<(), Error> {
+		each_tower_subfield!(
```
nikurt, line 177: Please add a comment. I have no idea what's going on.

crates/compute/src/cpu/layer.rs
BenjaminTrapani: I have removed the `From` constraint because it is redundant with the definition of `ExtensionField`. `T::B128` is an `ExtensionField` of all other tower levels in the tower family `T`. The definition still uses `from` because the conversion is attempting to upcast an existing subfield element instead of modifying the base of a packed field.

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,72 @@ use binius_field::{
...
 #[derive(Debug)]
 pub struct CpuExecutor;
 
 #[derive(Debug, Default)]
 pub struct CpuLayer<F: TowerFamily>(PhantomData<F>);
 
+/// evals is treated as a matrix with `1 << log_query_size` rows and each column is dot-producted
+/// with the corresponding query element. The result is written to the `output` slice of values.
+/// The evals slice may be any field extension defined by the tower family T.
+fn compute_left_fold<EvalType: TowerField, T: TowerFamily>(
```
jimpo, line 24: nit: Move function to follow the `impl` block below. It will make the file easier to read.

In general, impl blocks should usually immediately follow `struct` definitions.

crates/compute/src/layer.rs
```diff
@@ -122,6 +125,16 @@ pub trait ComputeLayer<F> {
 		coordinates: &[F],
 		data: &mut <Self::DevMem as ComputeMemory<F>>::FSliceMut<'_>,
 	) -> Result<(), Error>;
+
+	fn fold_left<'a>(
```
jimpo, line 129: Missing docstring, and I prefer the parameter names [in the original branch](https://github.com/IrreducibleOSS/binius/blob/jimpo/2025-03-25-hal-playtime/crates/hal/src/v2.rs#L177). Is there a reason you added parameters for the sizes rather than inspecting the slice lengths?

crates/compute/src/memory.rs
```diff
@@ -72,3 +72,25 @@ pub trait ComputeMemory<F> {
 		Self::split_at_mut(borrowed, mid)
 	}
 }
+
+pub struct FSliceWithTowerLevel<'a, F, Mem: ComputeMemory<F>> {
+	pub slice: Mem::FSlice<'a>,
+	pub tower_level: u8,
```
jimpo, line 78: nit: We use `usize` for tower levels everywhere, please use here as well.

crates/compute/src/memory.rs
```diff
@@ -72,3 +72,25 @@ pub trait ComputeMemory<F> {
 		Self::split_at_mut(borrowed, mid)
 	}
 }
+
+pub struct FSliceWithTowerLevel<'a, F, Mem: ComputeMemory<F>> {
```
jimpo, line 76: This needs a docstring

jimpo, line 76: Please rename to `SubfieldSlice`

crates/compute/src/cpu/layer.rs
```diff
@@ -7,14 +7,72 @@ use binius_field::{
...
+	memory::FSliceWithTowerLevel,
+};
 
 #[derive(Debug)]
 pub struct CpuExecutor;
 
 #[derive(Debug, Default)]
 pub struct CpuLayer<F: TowerFamily>(PhantomData<F>);
 
+/// evals is treated as a matrix with `1 << log_query_size` rows and each column is dot-producted
```
jimpo, line 21: See https://github.com/IrreducibleOSS/binius/blob/main/CONTRIBUTING.md#documentation for docstring formatting.

jimpo (changes requested):

crates/compute/src/layer.rs
BenjaminTrapani: The log len of `query` can be inferred from the slice len now that the tower level is guaranteed to be log2(128).

Regarding parameter names, I matched those in the Binius definition at https://github.com/IrreducibleOSS/binius/blob/main/crates/math/src/fold.rs#L132. Which names do yo prefer?

jimpo: I prefer the names I chose for the API in the linked branch. IMO they reflect what the operation does better, whereas the `binius_math::fold` parameter names match how it is used in the protocol code.

jimpo: Please import the parameter names and docstring from https://github.com/IrreducibleOSS/binius/blob/jimpo/2025-03-25-hal-playtime/crates/hal/src/v2.rs#L177

PR #294, opened 2025-05-02
[M3] Allow core constraint system to handle flushes with multiple selectors

djadjka:


crates/m3/src/builder/channel.rs
```diff
@@ -13,30 +13,30 @@ pub struct Flush {
...
+	pub selectors: Vec<ColumnIndex>,
 }
 
 /// Options for a channel flush.
 #[derive(Debug)]
 pub struct FlushOpts {
 	/// The number of times the values are flushed to the channel.
 	pub multiplicity: u32,
-	/// An optional reference to a column to select which values to flush.
+	/// An optional reference to a columns to select which values to flush.
```
jimpo, line 27:
```suggestion
	/// Selector columns that determine which row events are flushed.
```

crates/m3/src/builder/channel.rs
```diff
@@ -13,30 +13,30 @@ pub struct Flush {
 	pub direction: FlushDirection,
 	/// The number of times the values are flushed to the channel.
 	pub multiplicity: u32,
-	/// An optional reference to a column to select which values to flush.
+	/// An optional reference to a columns to select which values to flush.
```
jimpo, line 16:
```suggestion
	/// Selector columns that determine which row events are flushed.
```

crates/core/src/constraint_system/verify.rs
```diff
@@ -443,7 +406,42 @@ pub fn make_flush_oracles<F: TowerField>(
 										}
 										_ => None,
 									}),
-							)?,
+							)?
+					} else {
+						let offset = *permutation_challenge + const_linear_combination + F::ONE;
+						let arith_expr_linear = ArithExpr::Const(offset);
+						let var_offset = 1; // Var(0) represents the selector column.
```
jimpo, line 413: This looks like a mistake, should be `flush.selectors.len()` probably.

jimpo (changes requested):
Can you please write a test constraint system? This can go in `tests` in `binius_m3`.

jimpo (approved):

Issue #291, opened 2025-05-01
[evalcheck] Reduce CompositeMLE evaluations to MLEchecks


jimpo:
One of our MultilinearPolyVariants is `Composite`, which is a multilinear extension of a multilinear composite polynomial. This is handled in evalcheck as a [reduction to a sumcheck](https://github.com/IrreducibleOSS/binius/blob/e50d97ef84062f36e5b826fd053f584731cabacf/crates/core/src/protocols/evalcheck/verify.rs#L245). In greedy_evalcheck, this then gets [proven as a batch sumcheck](https://github.com/IrreducibleOSS/binius/blob/e50d97ef84062f36e5b826fd053f584731cabacf/crates/core/src/protocols/greedy_evalcheck/prove.rs#L80). The issue is that this gets proven using the `RegularSumcheckProver`, whereas `Composite` oracle evaluations should reduce to the optimized [EqIndSumcheckProver](https://github.com/IrreducibleOSS/binius/blob/be6981b22a458601c88fa42ae629fca3f30b2c30/crates/core/src/protocols/sumcheck/prove/eq_ind.rs#L376).

Diving into implementation details, evalcheck groups the reduced sumchecks using the `ConstraintSetBuilder` with a `ConstraintPredicate::Sum`. One way to approach this would be to add a new `ConstraintPredicate` variant, `ConstraintPredicate::Eval { eval: F, point: EvalPoint<F> }`. Then we would group these together and build the `EqIndSumcheckProver` instances accordingly.

PR #297, opened 2025-05-03
[Fix] Match the order of processing bivariate sumcheck claims in evalcheck

djadjka:
The verifier processes claims successively, while the prover processes all projected claims first, followed by all composite.
As a result, if a projected and a composite claim have the same n_vars, the order of sumcheck_claims will differ between the verifier and prover.

The fix is a bit ad hoc, but #291 will provide a cleaner long-term solution

### DIFF
--- a/crates/core/src/protocols/evalcheck/prove.rs
+++ b/crates/core/src/protocols/evalcheck/prove.rs
@@ -1,13 +1,13 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::{collections::HashSet, iter};
+use std::collections::HashSet;
 
 use binius_field::{PackedField, TowerField};
 use binius_hal::ComputationBackend;
 use binius_math::MultilinearExtension;
 use binius_maybe_rayon::prelude::*;
 use getset::{Getters, MutGetters};
-use itertools::{chain, izip};
+use itertools::chain;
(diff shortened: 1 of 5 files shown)


crates/core/src/protocols/evalcheck/prove.rs
```diff
@@ -60,9 +60,7 @@ where
 	// Internally used to collect subclaims without evaluations for future query and memoization
 	claims_without_evals: Vec<(MultilinearPolyOracle<F>, EvalPoint<F>)>,
 	// The list of claims that reduces to a bivariate sumcheck in a round.
-	projected_bivariate_claims: Vec<EvalcheckMultilinearClaim<F>>,
-	// The list of claims that reduces to a composite MLEcheck in a round.
-	composite_mle_claims: Vec<EvalcheckMultilinearClaim<F>>,
+	bivariate_claims: Vec<BivariateClaims<P::Scalar>>,
```
jimpo, line 63: "bivariate" is incorrectly named. The `Composite` sumcheck claims are not bivariate products compositions. You can just name this field `sumcheck_claims`.

jimpo (commented):
Can you please explain the bug more clearly in the PR description and add a unit case that fails for it?

jimpo (approved):

PR #299, opened 2025-05-05
[compute] Introduce `accumulate_kernels`

iradicek:


crates/compute/src/layer.rs
```diff
@@ -69,6 +92,21 @@ pub trait ComputeLayer<F> {
 		Ok((out1, out2))
 	}
 
+	/// Compiles an arithmetic expression to the evaluator.
+	fn compile_expr(&self, expr: &ArithExpr<F>) -> Result<Self::ExprEval, Error>;
+
+	// TODO: better docs
```
iradicek, line 98: Suggest docs?

crates/compute/src/layer.rs
```diff
@@ -122,6 +160,77 @@ pub trait ComputeLayer<F> {
 		coordinates: &[F],
 		data: &mut <Self::DevMem as ComputeMemory<F>>::FSliceMut<'_>,
 	) -> Result<(), Error>;
+
+	// TODO: better docs
```
iradicek, line 164: Suggest docs?

crates/compute/src/layer.rs
```diff
@@ -122,6 +160,77 @@ pub trait ComputeLayer<F> {
...
+	fn sum_composition_evals(
+		&self,
+		exec: &mut Self::KernelExec,
+		log_len: usize,
+		inputs: &[KernelBuffer<'_, F, Self::DevMem>],
```
iradicek, line 169: In the original version, this had the type `&[Self::FSlice<'_>]`. It seems this is the more appropriate type

crates/compute/src/layer.rs
```diff
@@ -122,6 +160,77 @@ pub trait ComputeLayer<F> {
...
+		composition: &Self::ExprEval,
+		batch_coeff: F,
+		accumulator: &mut Self::KernelValue,
+	) -> Result<(), Error>;
+}
+
+// TODO: Better docs
```
iradicek, line 176: Suggest docs?

crates/compute/src/layer.rs
```diff
@@ -122,6 +160,77 @@ pub trait ComputeLayer<F> {
...
+	ChunkedMut {
+		data: Mem::FSliceMut<'a>,
+		log_min_chunk_size: usize,
+	},
+	Local {
+		log_size: usize,
+	},
+}
+
+// TODO: better docs!
```
iradicek, line 191: Suggest docs?

jimpo:
I'll write docs and push

iradicek:
> I'll write docs and push

Thanks!

crates/compute/src/layer.rs
jimpo: I think we should stick with `Self::FSlice`. We can always cast an `FSliceMut` to an `FSlice` with `ComputeMemory::as_const`. Maybe we should even have a method that helps convert a `KernelBuffer` to an `FSlice` (passing through Ref and calling `as_const` on Mut).

iradicek: I reverted to `FSlice` and moved conversion to the test fn.

Additionally, I've added fns to `KernelBuffer` to extract `Ref` and `Mut` variants. Note that it's not quite possible (at least I don't see it right now) to directly convert `KernelBuffer::Mut` to `FSlice`: if we fant to return `FSlice` (in `Ref` case) we need to take ownership of `KernelBuffer`, but `Mem::as_const` wants a reference, so we cannot have a fn that does both conversions...

I did not fully understand the bit about casting `FSliceMut` to `FSlice` wrt. to this discussion.

crates/compute/src/cpu/layer.rs
```diff
@@ -124,6 +178,83 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
...
+					let log_size = checked_log_2(data.len());
+					let log_chunk_size = log_size - log_chunks;
+					KernelBuffer::Ref(<Self as ComputeLayer<T::B128>>::DevMem::slice(
+						data,
+						(i << log_chunk_size)..((i + 1) << log_chunk_size),
+					))
+				}
+				KernelMemMap::ChunkedMut { data, .. } => {
+					let log_size = checked_log_2(data.len());
+					let log_chunk_size = log_size - log_chunks;
```
nikurt, line 242: pedantic nit: Should this be `checked_sub()`?

crates/compute/src/cpu/layer.rs
```diff
@@ -124,6 +178,83 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
...
+					KernelBuffer::Mut(<Self as ComputeLayer<T::B128>>::DevMem::slice_mut(
+						data,
+						(i << log_chunk_size)..((i + 1) << log_chunk_size),
+					))
+				}
+				KernelMemMap::Local { log_size } => {
+					let log_chunk_size = *log_size - log_chunks;
```
nikurt, line 249: pedantic nit: Should this be `checked_sub()`?

crates/compute/src/layer.rs
```diff
@@ -1,18 +1,34 @@
...
+use binius_field::Field;
+use binius_math::ArithExpr;
+use binius_utils::checked_arithmetics::checked_log_2;
+
 use super::{alloc::Error as AllocError, memory::ComputeMemory};
+use crate::memory::SizedSlice;
 
 /// A hardware abstraction layer (HAL) for compute operations.
-pub trait ComputeLayer<F> {
+pub trait ComputeLayer<F: Field> {
```
nikurt, line 13: Please consider changing the type trait to `TowerTrait`, such that this interface doesn't have to care about multiple types of `Field`-like traits.

jimpo: Does the trait care? I'm not opposed, but I don't see the immediate need.

crates/compute/src/cpu/layer.rs
jimpo: Could be. Since this is a private method, I wrote this logic assuming `log_chunks` is chosen in the appropriate range (`KernelMemMap::log_chunks_range`) as a pre-condition.

This call will panic on underflow in debug mode.

crates/compute/src/layer.rs
jimpo: Pushed a commit to handle both cases.

crates/compute/src/layer.rs
iradicek: I'm also wondering how is that going to help with the trait.

crates/compute/src/cpu/layer.rs
iradicek: My reasoning: the caller ensures this is valid and this fn is private to a reference implementation used only in tests, so I guess it's not needed. I can also add it if others think it's needed.

crates/compute/src/layer.rs
iradicek: The `ComputeLayer` trait is defined only with reference to `Field` (no tower there), because it's needed for the `ArithExpr<F>` (otherwise it would bot be bounded at all). Unless I'm missing something.

The `CpuLayer` (binius-internal impl.) is parametrized by a `T: TowerFamily` and them implements `ComputeLayer<T::B128>`.

jimpo (approved):

PR #301, opened 2025-05-05
Report a lookup bug

sai-deng:


jimpo:
See https://github.com/IrreducibleOSS/binius/pull/302

PR #302, opened 2025-05-05
[m3] Surface error when lookup count is too high

jimpo:
Addresses https://github.com/IrreducibleOSS/binius/pull/301

### DIFF
--- a/crates/m3/src/gadgets/lookup.rs
+++ b/crates/m3/src/gadgets/lookup.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 Irreducible Inc.
 
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use binius_core::constraint_system::channel::ChannelId;
 use binius_field::{ExtensionField, PackedExtension, PackedField, PackedSubfield, TowerField};
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


crates/m3/src/gadgets/lookup.rs
```diff
@@ -212,4 +222,48 @@ mod tests {
...
+
+		let lookup_table_size = 64;
+		let mut rng = StdRng::seed_from_u64(0);
+		let values = repeat_with(|| B128::random(&mut rng))
+			.take(lookup_table_size)
+			.collect::<Vec<_>>();
+
+		let counts = vec![9; lookup_table_size];
+		let values_and_counts = iter::zip(values, counts)
+			.sorted_unstable_by_key(|&(_val, count)| Reverse(count))
```
djadjka, line 245: Not sure we need to sort a vec![9; lookup_table_size]

crates/m3/src/gadgets/lookup.rs
```diff
@@ -60,6 +60,16 @@ impl LookupProducer {
 		P: PackedExtension<B1>,
 		P::Scalar: TowerField,
 	{
+		if self.multiplicity_bits.len() < 32 {
```
djadjka, line 63: 32 looks like a magic number — it makes sense to move it into a const variable with a descriptive name and a comment.

PR #304, opened 2025-05-06
[M3] Automatic witness population for constants

Anex007:
Small MR to automatically populate with witness index for constants.


crates/m3/src/builder/table.rs
```diff
@@ -352,6 +352,7 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
 			namespaced_name,
 			ColumnDef::Constant {
 				poly: Arc::new(mle),
```
jimpo, line 354: Now that the data field is on here, the `poly` is probably unnecessary. It's fine to leave as is for now though.

jimpo (approved):

PR #314, opened 2025-05-07
Add FRI fold operation to ComputeLayer

GraDKh:


BenjaminTrapani (dismissed):

crates/compute/src/cpu/layer.rs
```diff
@@ -639,6 +756,76 @@ mod tests {
...
+		let mut data_in = compute.host_alloc(1 << (log_len + log_batch_size));
+		let data_in = data_in.as_mut();
+		for x_i in data_in.iter_mut() {
+			*x_i = <F as Field>::random(&mut rng);
+		}
+		let data_in = data_in.to_vec();
+
+		// Copy the buffer to device slice
+		let (mut data_in_slice, device_memory) =
+			C::DevMem::split_at_mut(device_memory, data_in.len());
```
BenjaminTrapani, line 784: The device FSlice allocations can be obtained using the bump allocator:

```
let device_allocator =
			BumpAllocator::<'a, F2, <C as ComputeLayer<F2>>::DevMem>::new(device_memory);
let mut data_in_slice = device_allocator.alloc(data_in.len()).unwrap();
```

crates/compute/src/cpu/layer.rs
```diff
@@ -22,6 +24,19 @@ pub struct CpuExecutor;
 #[derive(Debug, Default)]
 pub struct CpuLayer<F: TowerFamily>(PhantomData<F>);
 
+impl<T: TowerFamily> CpuLayer<T> {
```
jimpo, line 27: please put helper methods after the main trait `impl ComputeLayer` block.

crates/compute/src/cpu/layer.rs
```diff
@@ -224,6 +239,107 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
...
+			.zip(data_out.iter_mut())
+			.enumerate()
+		{
+			// Fill the values with the folded items
+			if log_batch_size == 0 {
+				values.iter_mut().zip(chunk.iter()).for_each(|(x_i, y_i)| {
+					*x_i = *y_i;
+				});
+			} else {
+				let folded_iter = chunk.chunks(1 << log_batch_size).map(|chunk| {
```
jimpo, line 303: We do this method with the tensor fold + inner product because it's somewhat easier with packed fields. However, in this setting of regular field elements, I think it'd be simpler to just do several rounds of 2-to-1 regular folding with linear extrapolations using `interleave_challenges`. In other words, instead of expanding the tensor at all, make the `values` scratchpad have length `1 << (log_len + log_batch_size)`. Copy the whole chunk in and then iterate `extrapolate_line_scalar` over pairs of adjacent elements at the `interleave_challenge`.

This is, I think, a better patter to follow in HW implementation. And it saves 1 multiplication out of 2^log_batch_size.

nikurt (dismissed):

crates/compute/src/cpu/layer.rs
GraDKh: Applied

jimpo (approved):

PR #316, opened 2025-05-07
Reproduce a bug with MUL gadget

sai-deng:


jimpo:
Thanks for the bug report, fixed in https://github.com/IrreducibleOSS/binius/pull/313!

PR #318, opened 2025-05-08
[m3] Test coverage & cleanups for multiplication gadget

jimpo:
Add unit test to catch issue with compilation bug. Also test cleanups in mul gadget.


djadjka (dismissed):
LGTM, but maybe it makes sense to add a comment in `populate` that we fill the exponents using ExpWitness

PR #321, opened 2025-05-08
Fix eval_expr

LindaGuiga:
This PR aims at fixing the `eval_expr` method. It also adds a unit test with columns with different numbers of elements per row.

### DIFF
--- a/crates/m3/src/builder/witness.rs
+++ b/crates/m3/src/builder/witness.rs
@@ -1069,14 +1069,12 @@ impl<'a, F: TowerField, P: PackedField<Scalar = F>> TableWitnessSegment<'a, P> {
 				})?;
 		let expr_circuit = ArithCircuit::from(expr.expr());
 
-		let all_used = expr_circuit.vars_usage();
 		let col_refs = partition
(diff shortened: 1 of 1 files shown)


crates/m3/src/builder/witness.rs
```diff
@@ -1068,18 +1068,22 @@ impl<'a, F: TowerField, P: PackedField<Scalar = F>> TableWitnessSegment<'a, P> {
...
+
+		let all_used = expr_circuit.vars_usage();
 		let col_refs = partition
 			.columns
 			.iter()
-			.zip(expr_circuit.vars_usage())
-			.map(|(col_index, used)| {
+			.enumerate()
+			.map(|(i, col_index)| {
+				let used = (i < all_used.len()) && all_used[i];
```
jimpo, line 1078: `zipping` with the `vars_usage` should be OK here, no? If it contains fewer entries than the # of partition columns, then those columns aren't used.

jimpo (commented):
Thanks for figuring this out!

pepyakin (approved):

PR #323, opened 2025-05-09
Serialize evalcheck proofs using u32s instead of bytes and usize

josephjohnston:
The recursive verifier works with words, not bytes or usize.
Currently evalcheck proofs are encoded using bytes and usize. Either we
- switch binius to encoding using words
- post-process binius proofs to convert from bytes and usize to words.

I propose the former, and this is the PR. The latter seems unnecessarily complex.


pepyakin (approved):

djadjka:
Do you think it makes sense to apply this directly in the transcript?

jimpo (approved):

PR #326, opened 2025-05-09
[rustfmt] Enforce maximum comment width

jimpo:


pepyakin (approved):
Nice, can support that! (Haven't looked over every changed file though)

PR #331, opened 2025-05-10
[m3] Indexed lookups with tallying

jimpo:
Indexed lookup tables are fixed-size tables where every entry is easily determined by its index.

Indexed lookup tables cover a large and useful class of tables, such as lookup tables for bitwise operations, addition of small integer values, multiplication, etc. The entry encodes an input and an output, where the index encodes the input. For example, a bitwise AND table would have 2 8-bit input values and one 8-bit output value. The index encodes the input by  concatenating the 8-bit inputs into a 16-bit unsigned integer.

Before a lookup table witness can be filled, the number of times each entry is read must be known. Reads from indexed lookup tables are a special case where the counts are difficult to track during emulation, because the use of the lookup tables is an arithmetization detail. For  example, emulation of the system model should not need to know whether integer additions within a table are constraint using zero constraints or a lookup table for the limbs. In most cases of practical interest, the lookup table is indexed.

The method to tally counts is to scan all tables in the constraint system and boundaries values, and identify those that pull from the lookup table's channel. Then we iterate over the values read from the table and count all the indices.


crates/m3/src/builder/indexed_lookup.rs
```diff
@@ -0,0 +1,431 @@
...
+///
+/// A vector of counts, whose length is equal to `1 << indexed_lookup.log_size()`.
+pub fn tally<P>(
+	cs: &ConstraintSystem<B128>,
+	// TODO: This doesn't actually need mutable access. But must of the WitnessIndex methods only
+	// allow mutable access.
+	witness: &mut WitnessIndex<P>,
+	boundaries: &[Boundary<B128>],
+	chan: ChannelId,
+	indexed_lookup: &impl IndexedLookup<B128>,
```
pepyakin, line 55: Feels like could be `dyn`.

crates/m3/src/builder/indexed_lookup.rs
```diff
@@ -0,0 +1,431 @@
...
+		+ PackedExtension<B128>,
+{
+	let mut counts = vec![0; 1 << indexed_lookup.log_size()];
+
+	// Tally counts from the tables
+	for table in &cs.tables {
+		for partition in table.partitions.values() {
+			for flush in &partition.flushes {
+				if flush.channel_id == chan && flush.direction == FlushDirection::Pull {
+					if let Some(table_index) = witness.get_table(table.id()) {
```
pepyakin, line 73: Why don't we hoist this check up, to be the first thing in the outer loop?

pepyakin, line 73: A table could be not initialized and this expression will return None, silently moving to the next one. Could that be a footgun for the user?

pepyakin (dismissed):
Love the docs!

crates/m3/src/builder/indexed_lookup.rs
```diff
@@ -0,0 +1,431 @@
...
+				let input_i = input[i];
+				let carry_in_bit = bool::from(get_packed_slice(&carry_in, i).val());
+
+				let (output_i, carry_out_bit) = input_i.overflowing_add(carry_in_bit.into());
+				output[i] = output_i;
+				set_packed_slice(&mut carry_out, i, B1::from(carry_out_bit));
+				merged[i] = ((carry_out_bit as u32) << 17)
+					| ((carry_in_bit as u32) << 16)
+					| ((output_i as u32) << 8)
+					| input_i as u32;
```
pepyakin, line 312: Wait innit this merge_incr_vals?

crates/m3/src/builder/indexed_lookup.rs
jimpo: Maybe. Is there a benefit to dyn > impl here?

crates/m3/src/builder/indexed_lookup.rs
jimpo: Good point

jimpo: Yeah, this assumes all the looker tables are filled before tally is called. I see your concern and am open to suggestions on fixing it. The bad case here would be that the code fills the looker table after. Maybe we need some way of "sealing" a table. Like having a call to indicate that a table, whether initialized or not, will not be modified afterwards?

crates/m3/src/builder/indexed_lookup.rs
jimpo: Yep, it is. This is fixed in the next change in the stack. OK to leave it as is in this one?

PR #346, opened 2025-05-13
[clippy]: re-enable `missing-const-for-fn` clippy lint

tcoratger:
After https://github.com/rust-lang/rust-clippy/issues/14020 has been solved


tcoratger:
Closing as long as `Vec::<T, A>::len` is not yet stable as a const fn, this is probably better since this generates a lot of exceptions to deal with

PR #351, opened 2025-05-14
[rust] Update from Rust 2021 to Rust 2024 edition

bergkvist:
Most of the file changes are due to rustfmt sorting dependencies differently in Rust 2024 edition.
- `rng.gen()` had to be changed to `rng.r#gen()` since gen is now a reserved keyword in Rust 2024.
- All unsafe functions must now wrap any unsafe code it calls in explicit unsafe blocks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 
 [workspace.package]
 version = "0.2.0"
-edition = "2021"
+edition = "2024"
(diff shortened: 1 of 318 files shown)


jimpo (approved):
Thanks!

Only comment is to change the `r#gen` hacks to the new [Rng::random](https://docs.rs/rand/latest/rand/trait.Rng.html#method.random) method. (See rng docs that gen is deprecated). Please do this as a follow-up.

PR #353, opened 2025-05-15
[core] miri

pepyakin:
This changeset is related to [CRY-341]. Specifically it patches some cheesy
spots but it also highlights the issue using miri.

To run the test you need to execute the command below:

    RUSTFLAGS="-C target-feature=-neon" cargo miri test -p binius_core -- batch_evaluate_add_mul

Disabling -neon is required for apple silicon hardware as the code tries to use
the SIMD optimized versions of field math. For intel hardware you might need
to do a similiar tweak and is left as an exercise for the reader.

Besides that, the stackalloc was also wrapped as `alloca` is not recognized
by miri.

[CRY-341]: https://linear.app/irreducible/issue/CRY-341/check-funny-business-in-arith-circuit

### DIFF
--- a/crates/core/src/polynomial/arith_circuit.rs
+++ b/crates/core/src/polynomial/arith_circuit.rs
@@ -1,6 +1,6 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-use std::{fmt::Debug, mem::MaybeUninit, sync::Arc};
+use std::{mem::MaybeUninit, sync::Arc};
 
 use binius_field::{ExtensionField, Field, PackedField, TowerField};
 use binius_math::{ArithCircuit, ArithCircuitStep, CompositionPoly, Error, RowsBatchRef};
@@ -8,7 +8,6 @@ use binius_utils::{
 	DeserializeBytes, SerializationError, SerializationMode, SerializeBytes, bail,
 	mem::{slice_assume_init_mut, slice_assume_init_ref},
 };
-use stackalloc::stackalloc_uninit;
 
 /// Convert the expression to a sequence of arithmetic operations that can be evaluated in sequence.
 fn convert_circuit_steps<F: Field>(
@@ -306,8 +305,7 @@ impl<F: TowerField, P: PackedField<Scalar: ExtensionField<F>>> CompositionPoly<P
 			}
 		}
 
-		// `stackalloc_uninit` throws a debug assert if `size` is 0, so set minimum of 1.
-		stackalloc_uninit::<P, _, _>(self.steps.len().max(1), |evals| {
+		alloc_scratch_space::<P, _, _>(self.steps.len(), |evals| {
 			let get_argument_value = |input: CircuitStepArgument<F>, evals: &[P]| match input {
 				// Safety: The index is guaranteed to be within bounds by the construction of the
 				// circuit
@@ -375,8 +373,7 @@ impl<F: TowerField, P: PackedField<Scalar: ExtensionField<F>>> CompositionPoly<P
 			});
 		}
 
-		// `stackalloc_uninit` throws a debug assert if `size` is 0, so set minimum of 1.
-		stackalloc_uninit::<P, (), _>((self.steps.len() * row_len).max(1), |sparse_evals| {
+		alloc_scratch_space::<P, (), _>(self.steps.len() * row_len, |sparse_evals| {
 			for (i, expr) in self.steps.iter().enumerate() {
 				let (before, current) = sparse_evals.split_at_mut(i * row_len);
 
@@ -437,12 +434,12 @@ impl<F: TowerField, P: PackedField<Scalar: ExtensionField<F>>> CompositionPoly<P
 						}
 					}
 					CircuitStep::AddMul(target, left, right) => {
-						let target = &before[row_len * target..(target + 1) * row_len];
+						let target = &mut before[row_len * target..(target + 1) * row_len];
 						// Safety: by construction of steps and evaluation order we know
 						// that `target` is not borrowed elsewhere.
 						let target: &mut [MaybeUninit<P>] = unsafe {
 							std::slice::from_raw_parts_mut(
-								target.as_ptr() as *mut MaybeUninit<P>,
+								target.as_mut_ptr() as *mut MaybeUninit<P>,
 								target.len(),
 							)
 						};
@@ -541,6 +538,32 @@ fn apply_binary_op<F: Field, P: PackedField<Scalar: ExtensionField<F>>>(
 				}
 			}
 		}
+	}
+}
+
+fn alloc_scratch_space<T, U, F>(size: usize, callback: F) -> U
+where
+	F: FnOnce(&mut [MaybeUninit<T>]) -> U,
+{
+	#[cfg(miri)]
+	{
+		use std::mem;
+		// We don't want to deal with running destructors.
+		assert!(!mem::needs_drop::<T>());
+		let mut scratch_space = Vec::<T>::with_capacity(size);
+		unsafe {
+			scratch_space.set_len(size);
+			let out =
+				callback(mem::transmute::<&mut [T], &mut [MaybeUninit<T>]>(scratch_space.as_mut()));
+			drop(scratch_space);
+			out
+		}
+	}
+	#[cfg(not(miri))]
+	{
+		// `stackalloc_uninit` throws a debug assert if `size` is 0, so set minimum of 1.
+		let size = size.max(1);
+		stackalloc::stackalloc_uninit(size, callback)
 	}
 }
 
@@ -789,6 +812,32 @@ mod tests {
 		CompositionPoly::batch_evaluate(&circuit, &batch_query.get_ref(), &mut batch_result)
 			.unwrap();
 		assert_eq!(&batch_result, &[expected1, expected2, expected3]);
+	}
+
+	#[test]
+	fn batch_evaluate_add_mul() {
+		// This test is focused on exposing the the currently present stacked borrows violation. It
+		// passes but still triggers `miri`.
+
+		type F = BinaryField8b;
+		type P = PackedBinaryField8x16b;
+
+		let expr = (ArithExpr::<F>::Var(0) * ArithExpr::Var(0))
+			+ (ArithExpr::Const(F::ONE) - ArithExpr::Var(0)) * ArithExpr::Var(0)
+			- ArithExpr::Var(0);
+		let circuit = ArithCircuitPoly::<F>::new(expr.into());
+
+		let typed_circuit: &dyn CompositionPoly<P> = &circuit;
+		assert_eq!(typed_circuit.binary_tower_level(), 0);
+		assert_eq!(typed_circuit.degree(), 2);
+		assert_eq!(typed_circuit.n_vars(), 1);
+
+		let mut evals = [P::default(); 1];
+		let batch_query = [[P::broadcast(F::new(1).into())]; 1];
+		let batch_query = RowsBatch::new_from_iter(batch_query.iter().map(|x| x.as_slice()), 1);
+		typed_circuit
+			.batch_evaluate(&batch_query.get_ref(), &mut evals)
+			.unwrap();
 	}
 
 	#[test]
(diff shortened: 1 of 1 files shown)


pepyakin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/353?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#353** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/353?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/353?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#352** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/352?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -536,6 +531,32 @@ fn apply_binary_op<F: Field, P: PackedField<Scalar: ExtensionField<F>>>(
...
+		}
+	}
+	#[cfg(not(miri))]
+	{
+		// `stackalloc_uninit` throws a debug assert if `size` is 0, so set minimum of 1.
+		let size = size.max(1);
+		stackalloc::stackalloc_uninit(size, callback)
+	}
+}
+
```
GraDKh, line 559: I would rather make an in-place wrapper with the same name and put it to utils (I assume we may have the same problem in other places).
Also we may get rid of the annoying panic when length is zero

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -429,12 +424,12 @@ impl<F: TowerField, P: PackedField<Scalar: ExtensionField<F>>> CompositionPoly<P
...
+						let target = &mut before[row_len * target..(target + 1) * row_len];
 						// Safety: by construction of steps and evaluation order we know
 						// that `target` is not borrowed elsewhere.
 						let target: &mut [MaybeUninit<P>] = unsafe {
 							std::slice::from_raw_parts_mut(
-								target.as_ptr() as *mut MaybeUninit<P>,
+								target.as_mut_ptr() as *mut MaybeUninit<P>,
 								target.len(),
 							)
 						};
```
GraDKh, line 435: This is a common use case. Maybe we could introduce a helper function in utils:
```
pub fn assume_uninit_slice_mut<T>(values: &mut [T]) -> &mut [T] {
    assert!(!std::mem::needs_drop::<T>());
   
   ...
}
```

That would not allow UB as lifetime is propagated here

GraDKh (changes requested):
Just several minor comments.

Do we have plans to add a job running miri to our CI?

pepyakin:
I am afraid this is not going to work. As of now, this test fails in miri as the code violates stacked borrows (which are a) work in progress aliasing model, and b) which miri checks). CRY-341 is meant to tackle that and make miri happy. As soon as it's done, though, it would be great to exercise our code (or subsets of it) with miri!

crates/core/src/polynomial/arith_circuit.rs
pepyakin: I was not able to find other places that use `stackalloc` in this way, though, there are other places in which its used.

Now, my gut feeling says that stackalloc does not give real advantage over using `Vec`. Exploring this is actually part of the task here (pt.3).

Now with all that said, do you still  think this is really a blocking issue? And if so, could you elaborate what do you mean by an in-place wrapper?

crates/core/src/polynomial/arith_circuit.rs
pepyakin: I've checked quickly and I was not able to find any other place like that. 

TBH, I am not sure if I agree, to me it's seems on contrary: that it's a niche case. Specifically, it seems to me that you don't need to conjure a `MaybeUninit` from a slice, which by definition cannot have uninitialized data (and if it had, then UB already and we cannot reason further about that execution).

crates/core/src/polynomial/arith_circuit.rs
GraDKh: Yes, you're right, now this is the only case for `stackalloc_uninit`, the other ones seems to be removed from the codebase.

>Now, my gut feeling says that stackalloc does not give real advantage over using Vec
In the synthetic tests there was a significant deffirence between `stackalloc` and `Vec`/`SmallVec`. Regarding this particular place possibly we should compare for the case of the most simple polynomial, e.g. `Var(0) + Va(1)` if there is any difference.

GraDKh: >And if so, could you elaborate what do you mean by an in-place wrapper

I just meant to move to the utils and rename to `stackalloc_uninit`

crates/core/src/polynomial/arith_circuit.rs
GraDKh: I think that I'm facing similar cases when implementing the fast HAL. HAL memory structures do not use `MaybeUninit` but some of our interfaces do for output data. For packed field/field slices this is totally legit to pass an initialized data to the place where it will be overwritten as `MaybeUninit`

pepyakin: Ah, ok, this is what I had in mind when I said "niche" and maybe it's not so niche after all. Thanks for the explaination.

pepyakin: I have tried to do that and bumped what it feels obvious now : )

We cannot really do this because that would require creation of two exclusive references on the same data, the very same problem that the issue OP is meant to resolve in the future.

However, I still submitted the utility for you convenience https://github.com/IrreducibleOSS/binius/pull/355

GraDKh: >  would require creation of two exclusive references on the same data

But their usage is mutually exclusive with the helper function, isn't it? While we are using the returned value the original one can't be accessed.

pepyakin: And that's exactly the crux of the problem. Below is a snipped version of the changes you requested.

```rust
// ①
let target = &mut before[row_len * target..(target + 1) * row_len];
// ②
let target: &mut [MaybeUninit<P>] =
	binius_utils::mem::slice_uninit_mut(target);
// ③
apply_binary_op(
	/* .. */
	before,
	target,
	/* .. */,
);
```

Here is the problem: at ① we create a create an exclusive reference of the slice `before` called T1. Then, at ② we create an exclusive slice T2 based on T1 which still references the original `before` slice.  Finally at ③ we are trying to pass both T1 and T2.

GraDKh: Oh, now I see, thanks! So the problem is not that there is some UB, but the code doesn't pass borrow checker

pepyakin: Correct! And to put everything together:

1. The original issue [CRY-341](https://linear.app/irreducible/issue/CRY-341/check-funny-business-in-arith-circuit) was meant to remove the UB.
2. This code as of now, in main, does exhibit UB, because the exclusive reference it creates here is NOT in fact exclusive. 
3. The PR presented here is not meant to solve this issue. It is basically me dumping code that I used to convince myself that there is indeed a violation of rules (at least stacked borrow ones).
4. If we do use `mem::slice_uninit_mut(target)` then the compiler will be able to rub this fact in our faces, which it did in the code snippet above. The borrow checker does not pass.
5. As a result of the CRY-341, I expect we would use a proper stack abstraction which would allow us to get rid of the slice/unint slice juggling and much of the unsafe code should be concentrated. Although as I haven't studied this code deeply this should not be taken as a promise, no refunds.

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -536,6 +531,32 @@ fn apply_binary_op<F: Field, P: PackedField<Scalar: ExtensionField<F>>>(
...
+	#[cfg(miri)]
+	{
+		use std::mem;
+		// We don't want to deal with running destructors.
+		assert!(!mem::needs_drop::<T>());
+		let mut scratch_space = Vec::<T>::with_capacity(size);
+		unsafe {
+			scratch_space.set_len(size);
+			let out =
+				callback(mem::transmute::<&mut [T], &mut [MaybeUninit<T>]>(scratch_space.as_mut()));
```
jimpo, line 547: Looks like this is the same as https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut

pepyakin: Bulls eye! I haven't thought about this...

I am going to risk it and apply the "Graphite-suggested fix"!

jimpo: The AI lied here. You need to move `set_len` to _after_ this line with `spare_capacity_mut()`.

pepyakin: Maybe 2026 is going to be the year of AI vibe coding...

pepyakin: Wait, why would I need `set_len` at all? We are not interested in the vec logic, only in it's backing storage/capacity...

jimpo: Right, good point. Removing `set_len` is enough then.

jimpo (approved):

crates/core/src/polynomial/arith_circuit.rs
```diff
@@ -544,6 +541,29 @@ fn apply_binary_op<F: Field, P: PackedField<Scalar: ExtensionField<F>>>(
 	}
 }
 
+fn alloc_scratch_space<T, U, F>(size: usize, callback: F) -> U
+where
+	F: FnOnce(&mut [MaybeUninit<T>]) -> U,
+{
+	use std::mem;
+	// We don't want to deal with running destructors.
+	assert!(!mem::needs_drop::<T>());
```
GraDKh, line 550: [nit] We are actually not dealing with destructors in this function, it is the responsibility of the `callback` to work safely with `MaybeUninit`

GraDKh (approved):

crates/core/src/polynomial/arith_circuit.rs
pepyakin: Oh... I think this does not matter too much, because this is really localized code and we can see the immediate T and the call site, and it's not really that relevant wrt safety without `set_len` etc, etc

But nonetheless I will take time to expand the answer, because I suspect there might be a misconception in your understanding. I apologize in advance if that comes across to be too excessive or too patronizing, but getting unsafe code right is important and in line with the golden rule I would've appreciated such an explaination myself. So with that out of the way...

First of all, there is no such thing as an unsafe closure, therefore either we allow `callback` to do anything you can do in the safe Rust or we mark `alloc_scratch_space`. I'll say in advance that I don't think this is much appropriate and therefore let's concentrate on the former.

One of the things you can do in Rust is panicking. IOW, any safe code panicking should not lead to violating safety invariants. Panicking in `callback` will unwind the stack and will lead to dropping the `Vec` and thus plural `T`. If `T` had a destructor **and** we `set_len` then we would be in a trouble because drop could have been called on uninitialized items. This could be solved by the "pre-pooping your pants" pattern (I kid you not it's a thing) or other careful engineering... 

But why would be bother, if the immediate use-case does not deal with `T: Drop`? I hope this explains the safety aspects and my reasoning why I went with it in the first place.

Now, even without `set_len` the problem still is present, but instead of being a safety concern it is transformed into leaking which is arguably much better than violation of safety, but still not correct. 

So in summary:

1. We very much do deal with the destructors in this function because should there be a panic it will have to be cleaned up in this function.
2. We cannot demand the closure to uphold the safety constraints, at least not without making the whole function `unsafe`, which IMO undesired.
3. And even if we could, the closure is not in an adequate place to uphold the safety, because it is only given MaybeUninit and having that the callee cannot really prevent destructors from running or anything like that...

PR #354, opened 2025-05-15
[compute] Test utils module for external testing

iradicek:
Extracts generic `layer` tests into `test_utils` module, under a feature, so that other interface implementers can also test agains this suite.


jimpo:
Can this be closed in favor of https://github.com/IrreducibleOSS/binius/pull/377?

PR #357, opened 2025-05-15
[m3] don't assume oracle id order in witness

pepyakin:
This changeset aims to fix the implicit dependency of the witness on the order
of oracles added in `ConstraintSystem::compile`.. See [CRY-356] for more
details.

This does several things:

1. Prefers using ColumnId over indices
2. Introduction of `OracleLookup`

This starts using the ColumnId where possible. This change should really make
much less likely to confuse indices and improve general type-safety, and as well
as simplify the code a bit (one type is better than multiple). However, this
makes the column IDs to carry table_id around, but I don't think this is a
problem since the code is not hot.

Introduction the `OracleLookup` type. Prior this changeset
the data about which oracle came from which columns was lost. This type
is meant to store this information explicitly to be read by the witness
handling code.

This opens up for other use-cases where you might want to pass some data
from the compilation process into the witness generation, as that seems to be
inherently important: compilation creates oracles from columns and witness
filling operates on oracles.

Ideally though IMO we should really introduce a special type that represents
the compilation output:

1. `TableStat` should be a function of the compiled constraint system. That
   said, we can easily account for that.
2. Warnings. Optionally catch silly things like a committed column that is not
   referenced in a virtual column or a constraint.
3. It will also contain the data similiar to OracleLookup.

[CRY-356]: https://linear.app/irreducible/issue/CRY-356/dont-assume-the-order-of-compilation-of-oracles

### DIFF
--- a/crates/m3/src/builder/channel.rs
+++ b/crates/m3/src/builder/channel.rs
@@ -2,13 +2,14 @@
 
 use binius_core::constraint_system::channel::{ChannelId, FlushDirection};
 
-use super::column::ColumnIndex;
+use super::ColumnId;
 use crate::builder::{B1, Col};
 
 /// A flushing rule within a table.
 #[derive(Debug)]
 pub struct Flush {
-	pub column_indices: Vec<ColumnIndex>,
+	// TODO: rename to columns.
+	pub column_indices: Vec<ColumnId>,
 	pub channel_id: ChannelId,
 	pub direction: FlushDirection,
 	/// The number of times the values are flushed to the channel.
@@ -16,7 +17,7 @@ pub struct Flush {
 	/// Selector columns that determine which row events are flushed
 	///
 	/// The referenced selector columns must hold 1-bit values.
-	pub selectors: Vec<ColumnIndex>,
+	pub selectors: Vec<ColumnId>,
 }
 
 /// Options for a channel flush.
--- a/crates/m3/src/builder/table.rs
+++ b/crates/m3/src/builder/table.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 Irreducible Inc.
 
-use std::sync::Arc;
+use std::{ops::Index, sync::Arc};
 
 use binius_core::{
 	constraint_system::channel::{ChannelId, FlushDirection},
@@ -20,7 +20,7 @@ use binius_utils::{
 };
 
 use super::{
-	B1, ColumnIndex, FlushOpts,
+	B1, ColumnIndex, ColumnPartitionIndex, FlushOpts,
 	channel::Flush,
 	column::{Col, ColumnDef, ColumnId, ColumnInfo, ColumnShape},
 	expr::{Expr, ZeroConstraint},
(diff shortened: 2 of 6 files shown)


crates/m3/src/builder/channel.rs
```diff
@@ -2,21 +2,22 @@
...
 
-use super::column::ColumnIndex;
+use super::ColumnId;
 use crate::builder::{B1, Col};
 
 /// A flushing rule within a table.
 #[derive(Debug)]
 pub struct Flush {
-	pub column_indices: Vec<ColumnIndex>,
+	// TODO: rename to columns.
```
pepyakin, line 11: To be addressed in a follow up.

crates/m3/src/builder/table.rs
```diff
@@ -420,13 +430,25 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
 	{
 		assert!(pow_bits.len() <= (1 << FExpBase::TOWER_LEVEL));
 
+		// TODO: Add check for F, FSub, VALUES_PER_ROW
```
pepyakin, line 433: This is just duplicating the comment as above. I may fix both of those as follow ups.

crates/m3/src/builder/witness.rs
```diff
@@ -1712,18 +1727,19 @@ mod tests {
 		let allocator = Bump::new();
 
 		let table_size = 123;
+		let statement = Statement {
+			boundaries: vec![],
+			table_sizes: vec![table_size],
+		};
+		let ccs = cs.compile(&statement).unwrap();
```
pepyakin, line 1734: To make it clear: now you have to compile the cs before you can call `into_multilinear_extension_index`. 

I should probably add a doc on that, but `into_multilinear_extension_index` didn't have a comment.

jimpo: Hmm, this hidden dependency seems error prone. Should `into_multilinear_extension_index()` require some object exposed by the compiled system? Or returned alongside it?

jimpo: Is that basically the "special type that represents the compilation output" that you discuss in the PR description?

jimpo (dismissed):

crates/m3/src/builder/witness.rs
pepyakin: > Is that basically the "special type that represents the compilation output" that you discuss in the PR description?

Yes, that, although there are multiple variations of it. In addition to the one you have in mind, it is also possible to plumb this object through the [WitnessIndex::new](https://github.com/IrreducibleOSS/binius/blob/7f77720b41a09e676a634549feb0fee4e19f8ff0/crates/m3/src/builder/witness.rs#L49-L66). This might be more future-proof should the `WitnessIndex` need to refer to the compilation output before the `into_multilinear_extension_index`.

> Hmm, this hidden dependency seems error prone. Should into_multilinear_extension_index() require some object exposed by the compiled system? Or returned alongside it?

Yes, I agree that it's not ideal, but I do think it's OK, hence this solution is presented here. To elaborate why, I think the logic of proving is straightforward in 99%. Basically it's always a variation of: build cs, compile cs, populate witness, into_multlinear_extension_index. I cannot imagine a situation where in production those steps could be conditional (besdies the normal fork between the verifier and the prover, but that's irrelevant in this case). And yeah, btw, petravm also follows this sequence. If you agree with this, then this panic will only affect the developer workflow, and it's not that this is the worst panic the dev will see 🙃 , it's going to pop up early in the development process, it's quick to fix, etc.

But yes, I think the ideal solution would be to return a compilation artifact object. I did not go for this approach from the get go simply because this PR is heavy as it is.  Lmk if you want it, I can build it in this stack or we can punt on this for now.

pepyakin: Done

jimpo (approved):

pepyakin:
### Merge activity

* **May 17, 10:11 AM EDT**: @pepyakin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/IrreducibleOSS/binius/357).

PR #358, opened 2025-05-15
[m3] Add privacy to ColumnId and Col

pepyakin:
This essentially hides ColumnIndex and ColumnPartitionIndex from the outside
world.

### DIFF
--- a/crates/m3/src/builder/column.rs
+++ b/crates/m3/src/builder/column.rs
@@ -10,11 +10,11 @@ use super::{structured::StructuredDynSize, table::TableId, types::B128};
 
 /// An index of a column within a table.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct ColumnIndex(pub(crate) usize);
+pub(crate) struct ColumnIndex(pub(crate) usize);
 
 /// An index of a column within a partition.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct ColumnPartitionIndex(pub(crate) usize);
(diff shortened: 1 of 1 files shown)


crates/m3/src/builder/column.rs
```diff
@@ -10,11 +10,11 @@ use super::{structured::StructuredDynSize, table::TableId, types::B128};
 
 /// An index of a column within a table.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct ColumnIndex(pub(crate) usize);
+pub(crate) struct ColumnIndex(pub(crate) usize);
```
jimpo, line 13: I have a distaste for `pub(crate)`. Could we instead make the builder submodules private in `mod.rs` and reexport only the things that should get public exposed?

pepyakin: Oh, so you mean that make a pass where we remove all `pub(crate)` and change the `builder/mod.rs` from 

```
pub use channel::*;
pub use column::*;
pub use constraint_system::*;
pub use error::*;
pub use expr::*;
pub use indexed_lookup::*;
pub use stat::*;
pub use statement::*;
pub use structured::StructuredDynSize;
pub use table::*;
pub use types::*;
pub use witness::*;
```

That exposes the items on the on-needed basis? I can do that. That sounds orthogonal to this PR since it seems to me it would need cleaning up many other places around m3, so we can either land this one or we can skip it (and I remove it from the stack). Your call.

jimpo: OK, let's land this and do that as a follow-up. :+1:

jimpo (approved):

jimpo (approved):

PR #367, opened 2025-05-16
Environment variable flag to dump transcript to a file

djadjka:
### TL;DR

Added functionality to dump proof transcripts to a file when the `BINIUS_DUMP_PROOF` environment variable is set.

### What changed?

- Modified the `finalize` method in `ProverTranscript` to check for the `BINIUS_DUMP_PROOF` environment variable
- When the environment variable is set, the proof transcript is written to the file specified by the variable's value
- Added necessary imports for file operations (`std::fs::File`, `io::Write`)
- Made a minor refactoring in `into_verifier` method to store the transcript in a variable before passing it to the constructor

### How to test?

1. Set the environment variable `BINIUS_DUMP_PROOF` to a file path:
   ```
   export BINIUS_DUMP_PROOF=/path/to/output/proof.bin
   ```
2. Run a process that generates a proof
3. Verify that the proof has been written to the specified file

### Why make this change?

This change enables easier debugging and analysis of proof transcripts by allowing them to be dumped to a file for inspection. This is particularly useful during development and testing of cryptographic protocols, as it allows for offline examination of the proof data.


djadjka:
* **#367** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/367?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/core/src/transcript/mod.rs
```diff
@@ -124,7 +126,15 @@ impl<Challenger_: Default + Challenger> Default for ProverTranscript<Challenger_
 
 impl<Challenger_: Challenger> ProverTranscript<Challenger_> {
 	pub fn finalize(self) -> Vec<u8> {
-		self.combined.buffer.to_vec()
+		let transcript = self.combined.buffer.to_vec();
```
djadjka, line 129: @jimpo you asked to add to `into_verifier`, but I moved it to the `finalize` instead to also have the ability to dump examples.

crates/core/src/transcript/mod.rs
```diff
@@ -124,7 +126,15 @@ impl<Challenger_: Default + Challenger> Default for ProverTranscript<Challenger_
...
+
+		if let Ok(filename) = std::env::var("BINIUS_DUMP_PROOF") {
+			let mut file = File::create(&filename)
+				.unwrap_or_else(|_| panic!("Failed to create proof dump file: {filename}"));
```
jimpo, line 133: nit: Use `expect(format(`

jimpo, line 133: nit: Use `expect(format(`

jimpo:
@benediamond Can you please test that this works for you?

benediamond:
ok, confirmed that i can now get binary files. when sitting in `/binius/` and executing `BINIUS_DUMP_PROOF=./ cargo test piop:: `, the files all land in `/crates/core`, instead of the top-level directory. that is pretty confusing behavior, seeing as i said `./`?

what if i run different tests? will they all land in `core`, or possibly somewhere else?

but this is probably good enough to merge now i suppose

benediamond (approved):

crates/core/src/transcript/mod.rs
```diff
@@ -124,7 +126,24 @@ impl<Challenger_: Default + Challenger> Default for ProverTranscript<Challenger_
 
 impl<Challenger_: Challenger> ProverTranscript<Challenger_> {
 	pub fn finalize(self) -> Vec<u8> {
-		self.combined.buffer.to_vec()
+		let transcript = self.combined.buffer.to_vec();
+
+		if let Ok(filename) = std::env::var("BINIUS_DUMP_PROOF") {
+			let path = if cfg!(test) {
+				let current_thread = std::thread::current();
```
jimpo, line 133: Please leave comments explaining what's going on here

GraDKh (commented):
Should we feature-gate this functionality in order to avoid problems with compiling againgst WebAssembly target?

djadjka:
I think we can add `#[cfg(not(target_arch = "wasm32"))]` for this case

crates/core/src/transcript/mod.rs
```diff
@@ -124,7 +126,35 @@ impl<Challenger_: Default + Challenger> Default for ProverTranscript<Challenger_
...
-		self.combined.buffer.to_vec()
+		let transcript = self.combined.buffer.to_vec();
+
+		// Dumps the transcript to the path set in the BINIUS_DUMP_PROOF env variable.
+		if let Ok(path) = std::env::var("BINIUS_DUMP_PROOF") {
+			let path = if cfg!(test) {
+				// Because tests may run simultaneously, each test includes its name in the file
+				// name to avoid collisions.
+				let current_thread = std::thread::current();
+				let test_name = current_thread.name().unwrap_or("unknown");
```
pepyakin, line 137: FWIW, dumps from multiple unknown threads will lead to name conflicts. Not blocking.

djadjka: This code runs only in tests, and it's hard to imagine a case where "unknown" could be returned — but I still think it's better than using `expect` here

pepyakin: Fair enough!

jimpo (approved):

jimpo:
### Merge activity

* **May 19, 9:50 AM EDT**: @jimpo merged this pull request with [Graphite](https://app.graphite.dev/github/pr/IrreducibleOSS/binius/367).

PR #369, opened 2025-05-16
Add fast CPU crate

GraDKh:


jimpo:
@GraDKh Can you please open this using Graphite?

PR #380, opened 2025-05-18
deps: rm unused deps

tcoratger:


GraDKh (approved):

tcoratger:
> Thanks! Do you have a tool to detect unused dependencies @tcoratger?

@jimpo Yep this one https://github.com/bnjbvr/cargo-machete but sometimes there are false positives so we should take care of them

PR #382, opened 2025-05-19
[compute] Fix docs & and use `SubfieldSlice` in `inner_product`

iradicek:


crates/compute/src/cpu/layer.rs
```diff
@@ -129,15 +129,15 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
 	fn inner_product<'a>(
 		&'a self,
 		_exec: &'a mut Self::Exec,
-		a_edeg: usize,
+		a_tower_height: usize,
 		a_in: &'a [T::B128],
```
GraDKh, line 133: I would personally switch to `SubfieldSlice` for `a_in`. Seems that that's the right case for that struct

iradicek: I'm fine with that. @jimpo any reason not to do that?

jimpo: Yeah, `SubfieldSlice` is better here. @BenjaminTrapani introduced it after I put the initial spec/proposal up.

iradicek: Done

GraDKh (approved):

PR #386, opened 2025-05-19
[m3] Binary merkle tree example internal nodes deduped. 

HiddenAndBound:
Deduped internal nodes in Binary Merkle tree example, addressed nits.

### DIFF
--- a/crates/m3/tests/merkle_tree.rs
+++ b/crates/m3/tests/merkle_tree.rs
@@ -3,7 +3,10 @@
 /// High-level model for binary Merkle trees using the Grøstl-256 output transformation as a 2-to-1
 /// compression function.
 mod model {
-	use std::{collections::HashSet, hash::Hash};
+	use std::{
+		collections::{HashMap, HashSet},
+		hash::Hash,
+	};
 
 	use binius_hash::groestl::{GroestlShortImpl, GroestlShortInternal};
 	use binius_m3::emulate::Channel;
@@ -33,7 +36,7 @@ mod model {
 		pub root_id: u8,
 		pub index: usize,
 		pub leaf: [u8; 32],
-		pub path: Vec<[u8; 32]>,
+		pub nodes: Vec<[u8; 32]>,
 	}
 
 	/// A struct whose fields contain the channels involved in the trace to verify merkle paths for
@@ -188,6 +191,11 @@ mod model {
 				index: self.parent_index,
 			});
 
+			println!(
+				"push (root_id: {:?}, depth: {:?}, index: {:?})",
+				self.root_id, self.parent_depth, self.parent_index
+			);
+
 			if self.flush_left {
 				node_channel.pull(NodeFlushToken {
 					root_id: self.root_id,
@@ -195,6 +203,12 @@ mod model {
 					depth: self.parent_depth + 1,
 					index: 2 * self.parent_index,
 				});
+				println!(
+					"pull (root_id: {:?}, depth: {:?}, index: {:?})",
+					self.root_id,
+					self.parent_depth + 1,
+					2 * self.parent_index
+				);
 			}
 			if self.flush_right {
 				node_channel.pull(NodeFlushToken {
@@ -203,6 +217,12 @@ mod model {
 					depth: self.parent_depth + 1,
 					index: 2 * self.parent_index + 1,
 				});
+				println!(
+					"pull (root_id: {:?}, depth: {:?}, index: {:?})",
+					self.root_id,
+					self.parent_depth + 1,
+					2 * self.parent_index + 1
+				);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


crates/m3/tests/merkle_tree.rs
```diff
@@ -152,24 +176,34 @@ mod model {
 	}
 
 	impl MerklePathEvent {
-		pub fn fire(&self, node_channel: &mut Channel<NodeFlushToken>) {
+		/// Method to fire the event, pushing the parent digest if the parent flag is set and
```
jimpo, line 179: What's the parent flag for?

HiddenAndBound: This is to ensure the parent is not pushed twice, when both child nodes exist across Merkle paths. In this case, the fire method would push the parent in both cases previously. The nodes have been deduped in the trace generation, but the parent node would still end up being pushed twice.  If the flag is toggled, the parent node will not be pushed.

crates/m3/tests/merkle_tree.rs
```diff
@@ -182,96 +215,193 @@ mod model {
...
 	}
 
 	/// Struct representing the trace of the merkle tree inclusion proof statement.
 	pub struct MerkleTreeTrace {
-		pub boundaries: Vec<MerkleBoundary>,
-		pub nodes: Vec<MerklePathEvent>,
-		pub root: Vec<MerkleRootEvent>,
+		pub boundaries: MerkleBoundaries,
+		pub nodes: HashSet<MerklePathEvent>,
+		pub root: HashSet<MerkleRootEvent>,
```
jimpo, line 258: Does this need to be a HashSet? Vec should be fine here.

jimpo (changes requested):
I think this can be simplified a bit. See inline comments.

crates/m3/tests/merkle_tree.rs
HiddenAndBound: I felt using hash sets allows for deduping naturally on inserts, because it automatically ignores insertions of duplicate entries.

crates/m3/tests/merkle_tree.rs
```diff
@@ -182,96 +215,193 @@ mod model {
...
-							right: *node,
-							parent,
-							parent_depth: path.len() - i - 1,
-							parent_index: index >> (i + 1),
-							flush_left: true,
-							flush_right: false,
-						});
+
+						if filled_nodes.contains(&(*root_id, path.len() - i, sibling_index)) {
+							path_nodes.remove(&MerklePathEvent {
```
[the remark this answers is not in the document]
HiddenAndBound: This would be much cleaner, I didn't think of this.

crates/m3/tests/merkle_tree.rs
HiddenAndBound: But after addressing the next comment, HashSets should not be necessary.

crates/m3/tests/merkle_tree.rs
```diff
@@ -182,96 +235,152 @@ mod model {
...
+				);
+
+				root_nodes.insert(MerkleRootEvent::new(*root_id, roots[*root_id as usize]));
+
+				let mut parent_node = [0u8; 32];
+				for (i, &node) in nodes.iter().enumerate() {
+					match filled_nodes.get_mut(&(*root_id, index >> (i + 1), nodes.len() - i - 1)) {
+						Some((_, _, _, flush_left, flush_right)) => {
+							if (index >> i) & 1 == 0 {
+								compress(&current_child, &node, &mut parent_node);
```
jimpo, line 326: You shouldn't need to recompress, you can read `parent_node` out of the `filled_nodes` value.

crates/m3/tests/merkle_tree.rs
```diff
@@ -152,24 +179,50 @@ mod model {
...
+				root_id: self.root_id,
+				data: self.parent,
+				depth: self.parent_depth,
+				index: self.parent_index,
+			});
+
+			println!(
+				"push (root_id: {:?}, depth: {:?}, index: {:?})",
+				self.root_id, self.parent_depth, self.parent_index
+			);
```
pepyakin, line 197: Are you planning on to leaving this and other println statements?

crates/m3/tests/merkle_tree.rs
```diff
@@ -182,96 +235,152 @@ mod model {
...
-							flush_right: true,
-						});
+			let mut path_nodes = Vec::new();
+			let mut root_nodes = HashSet::new();
+			let mut boundaries = MerkleBoundaries::new();
+			// Number of times each root is referenced in the paths. Since internal nodes have been
+			// deduped, these need to be pushed into the nodes channel as many times as they are
+			// referenced in the paths.
+
+			//tracks the filled nodes in the tree
```
pepyakin, line 292: Nit

```suggestion
			// Tracks the filled nodes in the tree
```

jimpo (approved):

PR #391, opened 2025-05-20
[compute] Introduce ComputeLayer::extrapolate_line operation

jimpo:
# Introduce ComputeLayer::extrapolate_line operation

This PR adds a new `extrapolate_line` operation to the `ComputeLayer` trait. This operation extrapolates a line between evaluations at 0 and evaluations at 1, computing `y_z = y_0 + (y_1 - y_0) * z` for a given point `z`. The operation is performed in parallel over vectors of field elements.

The implementation includes:
- Adding the `extrapolate_line` method to the `ComputeLayer` trait
- Implementing the method for `CpuLayer<T>`
- Adding test utilities and a test case to verify correctness
- Making `ComputeLayer<F: Field>` and `TowerFamily` require `'static`

The operation is useful for interpolating between two points on a line and evaluating at arbitrary positions.

PR #389, opened 2025-05-19
[CRY-364] Add bivariate_product round calculation function

jimpo:
This PR adds a new `v3` module to the sumcheck protocol implementation, which contains a specialized function for calculating round evaluations for bivariate product compositions. The implementation leverages the `binius_compute` crate for hardware acceleration.

Key changes:
- Add new `calculate_round_evals` function that computes batched evaluations of products of pairs of partially specialized multilinear polynomials
- Re-export `layer` and `memory` modules from the compute crate
- Add `binius_compute` dependency to the core crate
- Add getters for the `indices` field in `IndexComposition`

This is the first step toward a V3 sumcheck prover implementation that will eventually replace the current V2 prover.

PR #392, opened 2025-05-20
[compute] Fix allocation lifetime issue

jimpo:
Objects allocated by an allocator must not outlive the allocator. This precludes allocators from owning data and allocating parts of it. The new interface matches the bumpalo [alloc](https://docs.rs/bumpalo/latest/bumpalo/struct.Bump.html#method.alloc) method now.

### DIFF
--- a/crates/compute/src/memory.rs
+++ b/crates/compute/src/memory.rs
@@ -35,6 +35,9 @@ pub trait ComputeMemory<F> {
 	/// Borrows an immutable memory slice, narrowing the lifetime.
 	fn narrow<'a>(data: &'a Self::FSlice<'_>) -> Self::FSlice<'a>;
 
+	/// Borrows a mutable memory slice, narrowing the lifetime.
+	fn narrow_mut<'a, 'b: 'a>(data: Self::FSliceMut<'b>) -> Self::FSliceMut<'a>;
+
 	/// Borrows a mutable memory slice as immutable.
 	///
 	/// This allows the immutable reference to be copied.
--- a/crates/compute/src/alloc.rs
+++ b/crates/compute/src/alloc.rs
@@ -5,7 +5,7 @@ use std::cell::Cell;
 use super::memory::{ComputeMemory, SizedSlice};
 use crate::cpu::CpuMemory;
 
-pub trait ComputeAllocator<'a, F, Mem: ComputeMemory<F>> {
+pub trait ComputeAllocator<F, Mem: ComputeMemory<F>> {
 	/// Allocates a slice of elements.
 	///
 	/// This method operates on an immutable self reference so that multiple allocator references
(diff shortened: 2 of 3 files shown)


jimpo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/392?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#393** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/393?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#392** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/392?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/392?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#391** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/391?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#389** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/389?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#388** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/388?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

iradicek:
I wonder, does this have to do with problem of passing mutable ref slice into kernel-level executor, e.g., `accumulate_kernels`?

iradicek (approved):

jimpo:
> s this have to do with proble

Mmm, I don't see a direct connection, but the memory allocator lifetime issue is pretty fundamental.

crates/compute/src/memory.rs
```diff
@@ -35,6 +35,9 @@ pub trait ComputeMemory<F> {
 	/// Borrows an immutable memory slice, narrowing the lifetime.
 	fn narrow<'a>(data: &'a Self::FSlice<'_>) -> Self::FSlice<'a>;
 
+	/// Borrows a mutable memory slice, narrowing the lifetime.
+	fn narrow_mut<'a, 'b: 'a>(data: Self::FSliceMut<'b>) -> Self::FSliceMut<'a>;
```
GraDKh, line 39: Maybe it would be better to borrow instead of consuming the original value.
so that we can do the following:
```
let mut original_data = ...;
some_method(Mem::narrow(&mut original_data));
// continiue using original_data here

Comment (GraDKh) [inline crates/compute/src/memory.rs:39]: Maybe it would be better to borrow instead of consuming the original value.
so that we can do the following:
```
let mut original_data = ...;
some_method(Mem::narrow(&mut original_data));
// continiue using original_data here

GraDKh (approved):

crates/compute/src/memory.rs
jimpo: Unfortunately, I think we need this signature that consumes the value as well. I couldn't figure another way to make it work in the place above. :=/

GraDKh: Yeah, maybe we need to duplicate some of the methods to work both with mutable references and temporary values

PR #394, opened 2025-05-20
AVX2 Parallel 4-instance Groestl Hasher 

jkirmayer:
Byte-sliced hasing of 4 groestl instances in parallel on AVX2 machines.

PR #397, opened 2025-05-20
[hash] Benchmarks and fallback implementation for AVX2 parallel Groestl

jkirmayer:
Benchmarks for Groestl256Multi and a MultiDigest impl for a wrapper on all types implementing Digest


crates/hash/src/groestl/arch/groestl_multi_avx2.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+#[inline]
+#[allow(clippy::identity_op)]
+fn shift_bytes_q(state: &mut State) {
+	state[0] = unsafe { _mm256_ror_epi64(state[0], 8 * 1) };
+	state[1] = unsafe { _mm256_ror_epi64(state[1], 8 * 3) };
+	state[2] = unsafe { _mm256_ror_epi64(state[2], 8 * 5) };
+	state[3] = unsafe { _mm256_ror_epi64(state[3], 8 * 7) };
+	// state[4] = unsafe { _mm256_ror_epi64(state[4], 8 * 0) };
```
pepyakin, line 184: eh? one above state[0] as well.

Does it work without this?

jkirmayer: I thought I'd leave these comments in to show the full shift-bytes step without the zero ops, but I can remove them if that's cleaner

jkirmayer: I think you're right about this, and whats going on here can pretty easily be deduced. Ended up just removing it

crates/hash/src/multi_digest.rs
```diff
@@ -172,6 +173,60 @@ impl<D: Digest + Send + Sync + Clone> ParallelDigest for D {
...
+	for MultipleDigests<D, N>
+{
+	type Digest = D;
+
+	fn new() -> Self {
+		Self(array::from_fn(|_| D::new()))
+	}
+
+	fn update(&mut self, data: [&[u8]; N]) {
+		self.0.par_iter_mut().enumerate().for_each(|(i, hasher)| {
```
GraDKh, line 189: I personally would not use parallel iterator inside this implementation, because "normal" implementations of `MultiDigest` are single-threaded and we have a level of parallelization somewhere on other levels of the call-stack

crates/hash/src/multi_digest.rs
```diff
@@ -182,6 +237,7 @@ mod tests {
 	use rand::{RngCore, SeedableRng, rngs::StdRng};
 
 	use super::*;
+	use crate::groestl::Groestl256;
 
 	#[derive(Clone, Default)]
 	struct MockDigest {
```
GraDKh, line 243: Seems like we can remove the this one and use `MultipleDigests` instead

GraDKh (changes requested):

GraDKh (approved):

GraDKh (changes requested):
Everything looks good but now I see the changes from https://github.com/IrreducibleOSS/binius/pull/394/
@jkirmayer did you use Graphite for this one?

jkirmayer:
I ended up doing some manual resetting after struggling with graphite and rebasing onto the first PR

jimpo:
Base change was merged, please rebase onto main

crates/hash/src/multi_digest.rs
```diff
@@ -172,16 +172,71 @@ impl<D: Digest + Send + Sync + Clone> ParallelDigest for D {
 	}
 }
 
+#[derive(Clone)]
+pub struct MultipleDigests<D: Digest, const N: usize>([D; N]);
```
jimpo, line 176: Do we need this? `ParallelDigest` is sufficient for merkle tree integration. And for the purposes of benchmarking, we can compare regular digest throughput to Multidigest throughput.

crates/hash/src/groestl/arch/groestl_multi_avx2.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2025 Irreducible Inc.
 
 use std::{arch::x86_64::*, array};
+
+use crate::multi_digest::ParallelMulidigestImpl;
```
jimpo, line 5: nit: typo

jimpo: You should install the pre-commit hooks, which would detect this. `pre-commit install`

GraDKh (dismissed):

GraDKh (approved):

PR #399, opened 2025-05-20
[compute] Use ComputeLayer::map for bivariate products

jkirmayer:
### TL;DR

Added a new `map` method to the `ComputeLayer` trait to support concurrent execution of operations.

### What changed?

- Added a corresponding test in the compute crate that uses the new map function with multilinear evaluations
- Updated the bivariate product implementation to use the new map method instead of sequential processing

### Why make this change?

This change enables more efficient parallel processing of operations by providing a higher-level abstraction for mapping operations across multiple inputs. It allows the compute layer implementations to potentially optimize concurrent execution of similar operations, which can lead to better performance, especially on hardware that supports parallelism.


crates/core/src/protocols/sumcheck/v3/bivariate_product.rs
```diff
@@ -161,11 +161,9 @@
 
 		// Fold the multilinears
 		let _ = self.hal.execute(|exec| {
-			// TODO: Parallelize hal operations with hal.map
 			self.multilins = self
-				.multilins
-				.drain(..)
-				.map(|multilin| {
+				.hal
+				.map(exec, self.multilins.drain(..), |exec, multilin| {
```
jimpo, line 166: Split this into a separate PR in the stack. You can try `gt split`

jimpo:
Please edit the PR name to be more descriptive, as it becomes the commit message. And prefix the PR name with `[compute]` to indicate that it's a change to the compute module, per repo convention.

jimpo (approved):

PR #402, opened 2025-05-20
[compute] Fix swapping single device element by copying the whole buffer

iradicek:


crates/compute_test_utils/src/layer.rs
```diff
@@ -164,10 +164,19 @@ pub fn test_generic_multiple_multilinear_evaluations<
 	let results = compute
 		.execute(|exec| {
 			{
-				// Swap first element on the device buffer
-				let mut first_elt =
-					<C::DevMem as ComputeMemory<F>>::slice_mut(&mut eq_ind_slice, ..1);
-				compute.copy_h2d(&[F::ONE], &mut first_elt)?;
+				// Swap first element (by copying the whole slice)
+				let mut tmp_eq_ind_buffer = compute.host_alloc(1 << n_vars);
```
GraDKh, line 168: You don't need to copy the whole slice, just `ComputeMemory::MIN_SLICE_LEN`, modify the first lement and copy it back

GraDKh (approved):

crates/compute_test_utils/src/layer.rs
iradicek: `MIN_SLICE_LEN` will be (or already is) `1` (this is minimal slice size on device) and does not apply here. Host to device and vice-versa copies are different, unfortunately.

iradicek: I think I will close this and fix it on HW side.

PR #409, opened 2025-05-21
Implement `FastCpuLayer::accumulate_kernels`

GraDKh:


djadjka (approved):

crates/compute/src/cpu/layer.rs
```diff
@@ -389,25 +386,14 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
 	}
 }
 
+unsafe impl<T: TowerFamily> Sync for CpuLayer<T> {}
```
jimpo, line 389: Why is this an `unsafe impl`? What's not `Sync` about `CpuLayer` by default?

crates/fast_compute/src/layer.rs
```diff
@@ -223,13 +282,46 @@ impl<T: TowerFamily, P: PackedTop<T>> ComputeLayer<T::B128> for FastCpuLayer<T,
...
-		_composition: &Self::ExprEval,
-		_batch_coeff: T::B128,
-		_accumulator: &mut T::B128,
+		inputs: &SlicesBatch<FSlice<'_, T::B128, Self>>,
+		composition: &Self::ExprEval,
+		batch_coeff: T::B128,
+		accumulator: &mut Self::KernelValue,
 	) -> Result<(), Error> {
-		unimplemented!()
+		const BATCH_SIZE: usize = 64;
```
jimpo, line 290: Please leave a comment explaining this choice. Even if it's arbitrary, that should go in the comment.

crates/compute/src/memory.rs
```diff
@@ -22,12 +24,49 @@ impl<T> SizedSlice for &mut [T] {
...
 
+/// A batch of slices of the same length.
+pub struct SlicesBatch<Slice: SizedSlice> {
+	rows: Vec<Slice>,
+	row_len: usize,
+}
+
+impl<Slice: SizedSlice> SlicesBatch<Slice> {
+	/// Creates a new batch of slices with the given length
+	pub fn new(rows: Vec<Slice>, row_len: usize) -> Self {
```
jimpo, line 35: Please document the pre-condition in the docstring. Otherwise it should return an `InputValidation` error instead of asserting.

crates/compute/src/layer.rs
```diff
@@ -299,8 +301,7 @@ pub trait ComputeLayer<F: Field>: 'static {
 	fn sum_composition_evals(
 		&self,
 		exec: &mut Self::KernelExec,
-		log_len: usize,
-		inputs: &[FSlice<'_, F, Self>],
+		inputs: &SlicesBatch<FSlice<'_, F, Self>>,
```
jimpo, line 304: @BenjaminTrapani @iradicek Please note the interface change here

crates/fast_compute/src/layer.rs
```diff
@@ -190,14 +198,65 @@ impl<T: TowerFamily, P: PackedTop<T>> ComputeLayer<T::B128> for FastCpuLayer<T,
...
 			Vec<KernelBuffer<'a, T::B128, Self::DevMem>>,
 		) -> Result<Vec<Self::KernelValue>, Error>,
-		_mem_maps: Vec<KernelMemMap<'_, T::B128, Self::DevMem>>,
+		mem_maps: Vec<KernelMemMap<'_, T::B128, Self::DevMem>>,
 	) -> Result<Vec<Self::OpValue>, Error> {
-		unimplemented!()
+		let log_chunks_range = KernelMemMap::log_chunks_range(&mem_maps)
+			.ok_or_else(|| Error::InputValidation("no chunks range found".to_string()))?;
+
+		// Choose the number of chnks based on the range and the number of threads available.
```
jimpo, line 212: nit: chnks -> chunks

There's a pre-commit hook to catch typos now. Please install that or start running it regularly. (Or add to CI :-) )

jimpo (commented):
@BenjaminTrapani @iradicek Can you please review that the changes to the `binius_compute` crate are OK for you? (No need to review the `binius_fast_compute` code unless you want to)

crates/compute/src/layer.rs
```diff
@@ -444,6 +445,41 @@ impl<'a, F, Mem: ComputeMemory<F>> KernelMemMap<'a, F, Mem> {
...
+}
+
+/// A single chunk from [`KernelMemMap`].
+pub enum KernelMemMapChunk<'a, F, Mem: ComputeMemory<F>> {
+	Slice(Mem::FSlice<'a>),
+	SliceMut(Mem::FSliceMut<'a>),
+	Local(usize),
+}
+
+impl<'a, F, Mem: ComputeMemory<F>> Default for KernelMemMapChunk<'a, F, Mem> {
```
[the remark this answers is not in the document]
iradicek: Don't we already have a similar type?

crates/compute/src/layer.rs
iradicek: I say this because earlier this was a slice of column-slices and now `SlicesBatch` takes `rows`

crates/compute/src/cpu/layer.rs
GraDKh: This is because of `PhantomData<T>` requires T to be `Sync`

crates/compute/src/layer.rs
GraDKh: > Hm, it seems we cannot access individual column anymore.

@iradicek Actually we can via `SlicesBatch::row` method. The major difference between `&[FSlice<'_, F, Self>]` and `&SlicesBatch<FSlice<'_, F, Self>>` is that for the latter we don't need to check that all slices are of the same length inside the function which may save quite a signifficant time when we re-use the same big batch in many calls each of them accessing only a a few rows (that is a real case we met in CPU implementations)

crates/compute/src/cpu/layer.rs
```diff
@@ -93,7 +93,8 @@ impl<T: TowerFamily> ComputeLayer<T::B128> for CpuLayer<T> {
 	fn accumulate_kernels(
 		&self,
 		_exec: &mut Self::Exec,
-		map: impl for<'a> Fn(
+		map: impl Sync
```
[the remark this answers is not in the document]
GraDKh: If we want execute the code for different kernels in parallels I don't think there is a way to avoid this :(
We need this closure to be shared between threads

crates/compute/src/layer.rs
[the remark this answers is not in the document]
GraDKh: I've revised this with a fresh eye and agree it can be easily removed

crates/compute/src/cpu/layer.rs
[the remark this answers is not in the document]
jimpo: Yeah, unfortunately I think you're right. We can go ahead with this for now but will need to give more thought to a proper solution that matches the HW impl soon.

PR #420, opened 2025-05-22
[compute] Add Map to ComputeLayer

jkirmayer:


crates/compute/src/layer.rs
```diff
@@ -97,6 +97,16 @@ pub trait ComputeLayer<F: Field>: 'static {
 		Ok((out1, out2))
 	}
 
+	/// Creates an operation that depends on the concurrent execution of a sequence of operations.
+	fn map<Out, I: ExactSizeIterator>(
+		&self,
+		exec: &mut Self::Exec,
+		iter: I,
```
GraDKh, line 104: Just in case: the only way to parallelize it with the current signature in the fast CPU implementation is to produce a new rayon task for every item produced by iterator. 

So I assume the cost of the `map` call is supposed to be relatively big?

GraDKh (approved):

crates/compute/src/layer.rs
jkirmayer: Yes, I would think the function inside the map should make the cost of the new rayon task look small by comparison

PR #444, opened 2025-05-23
Stacked Allocation Pattern for ComputeLayer memory

jkirmayer:
Added ```remaining``` method to ComputeLayer allocators to create stack frames of ComputeLayer memory that are freed once the memory returned by ```remaining``` goes out of scope.

### DIFF
--- a/crates/compute/src/alloc.rs
+++ b/crates/compute/src/alloc.rs
@@ -125,24 +125,17 @@ mod tests {
 	fn test_stack_alloc() {
 		let mut data = (0..256u128).collect::<Vec<_>>();
 		let mut bump = BumpAllocator::<u128, CpuMemory>::new(&mut data);
-		println!("got here");
 		assert_eq!(bump.alloc(100).unwrap().len(), 100);
-		println!("got here");
-
 		assert_matches!(bump.alloc(200), Err(Error::OutOfMemory));
 
 		{
 			let next_layer_memory = bump.remaining();
 			let bump2 = BumpAllocator::<u128, CpuMemory>::new(next_layer_memory);
-			let _ = bump2.alloc(100);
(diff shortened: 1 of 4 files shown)


crates/compute/src/alloc.rs
```diff
@@ -17,6 +17,12 @@ pub trait ComputeAllocator<F, Mem: ComputeMemory<F>> {
 	/// - `n` must be a multiple of `Mem::MIN_SLICE_LEN`
 	fn alloc(&self, n: usize) -> Result<Mem::FSliceMut<'_>, Error>;
 
+	/// Allocates the remaining elements in the slice
+	///
+	/// This allows another allocator to have unique mutable access to the rest of the elements in
+	/// this allocator until it gets dropped, at which point this allocator can be used again
+	fn remaining(&mut self) -> &mut Mem::FSliceMut<'a>;
```
jimpo, line 24: This should not return an `'a` lifetime, it needs to return `'_`. That is a fancy way of doing:

```rust
fn remaining<'a>(&'a mut self) -> &mut Mem::FSliceMut<'a>;
```

jimpo (changes requested):
Please use Graphite :-)

crates/compute/src/alloc.rs
```diff
@@ -5,7 +5,7 @@ use std::cell::Cell;
 use super::memory::{ComputeMemory, SizedSlice};
 use crate::cpu::CpuMemory;
 
-pub trait ComputeAllocator<F, Mem: ComputeMemory<F>> {
+pub trait ComputeAllocator<'a, F, Mem: ComputeMemory<F>> {
```
jimpo, line 8: Remove 'a

jimpo (approved):

PR #458, opened 2025-05-26
[compute] Workaround for host allocation issue in bivariate sumcheck test

iradicek:
Host memory allocation on the HW Compute Layer currently fails if the requested allocation size is less than 512 elements. As a temporary workaround, the test for bivariate sumcheck has been modified to always allocate at least 512 words, even if fewer are actually needed. This change can be removed once the HW Compute Layer no longer has this limitation.

### DIFF
--- a/crates/compute_test_utils/src/bivariate_sumcheck.rs
+++ b/crates/compute_test_utils/src/bivariate_sumcheck.rs
@@ -182,10 +182,13 @@ pub fn generic_test_bivariate_sumcheck_prove_verify<F, Hal>(
 	)
 	.unwrap();
 
-	let mut host_mem =
-		hal.host_alloc(<BivariateSumcheckProver<F, Hal>>::required_host_memory(&claim));
+	// TODO[SYS-275]: Remove this once we can allocate less than 512 words on host in HW.
+	const MIN_DEVICE_TRANSFER_SIZE: usize = 1 << 9;
+	let claim_req_mem = <BivariateSumcheckProver<F, Hal>>::required_host_memory(&claim);
+	let mut host_mem = hal.host_alloc(std::cmp::max(claim_req_mem, MIN_DEVICE_TRANSFER_SIZE));
+	let host_mem = &mut host_mem.as_mut()[..claim_req_mem];
 
-	let host_alloc = HostBumpAllocator::new(host_mem.as_mut());
+	let host_alloc = HostBumpAllocator::new(host_mem);
 	let dev_alloc = BumpAllocator::new(dev_mem);
 
 	let dev_multilins = evals


crates/compute_test_utils/src/bivariate_sumcheck.rs
```diff
@@ -182,10 +182,12 @@ pub fn generic_test_bivariate_sumcheck_prove_verify<F, Hal>(
 	)
 	.unwrap();
 
-	let mut host_mem =
-		hal.host_alloc(<BivariateSumcheckProver<F, Hal>>::required_host_memory(&claim));
+	const MIN_DEVICE_TRANSFER_SIZE: usize = 1 << 9; // TODO: This needs to be exposed somewhere
```
jimpo, line 185: @BenjaminTrapani You suggested that we should remove `host_alloc` from `ComputeLayer` entirely, right?

iradicek: How can we do that? For HW version we need to allocate mapped memory, right?

BenjaminTrapani: It seems best for the HAL creation function to return a pair of (host_allocator, device_allocator). Supporting `host_alloc` on the HAL requires that all allocations performed with this API are preserved for the duration of the HAL, which is seldom the desired functionality in compute operations.

To support this API in tandem with stack-local temporary host allocations, we would need to partition the host allocator into two fixed width blocks, one for long-lived allocations via `host_alloc` and another for stack-local bump allocations. It may be simplest for implementations to provide complete allocators for host and device memory to Binius on construction. This way, Binius can partition these allocators as finely as needed for the scope of different allocations, continuing to reserve a section for completely persistent allocations if desired. 

Hardware-accelerated implementations have an opportunity to split part of both allocators for internal use during HAL construction (either cached artifacts across kernel executions or temporary scratch space reused per kernel). The HAL creation functions should accept both allocators and should return a split component of each input allocator for use by Binius.

crates/compute_test_utils/src/bivariate_sumcheck.rs
```diff
@@ -182,10 +182,12 @@ pub fn generic_test_bivariate_sumcheck_prove_verify<F, Hal>(
...
+	let claim_req_mem = <BivariateSumcheckProver<F, Hal>>::required_host_memory(&claim);
+	let mut host_mem = hal.host_alloc(std::cmp::max(claim_req_mem, MIN_DEVICE_TRANSFER_SIZE));
```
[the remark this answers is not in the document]
BenjaminTrapani: What do you think about requiring that hardware-accelerated implementations always return the requested slice length? If the allocator over-allocates for alignment or padding required by the device, this is an implementation detail and does not need to be exposed to Binius. For our implementation, we can implement the Binius allocator trait for our host allocators, trimming the allocated slice to the desired length for the Binius allocator implementation only.

GraDKh (dismissed):

jimpo:
Please add a PR description. This looks very hacky and I don't understand why it's necessary.

BenjaminTrapani (approved):

PR #461, opened 2025-05-26
[core] Introduce structured oracle type

pepyakin:
Structured is an oracle type that is very similar to Transparent, but is
defined by an ArithCircuit of some upper bound of n_vars.

This oracle type is mainly motivated by the needs of structured columns. The
idea is that such a structured column will define an arith expression up to some
maximum size. The frontend can supply any table size up to that maximum.

See [CRY-396] for more details.

[CRY-396]: https://linear.app/irreducible/issue/CRY-396/transparent-oracle-type-based-on-arithcircuit

### DIFF
--- a/crates/core/src/oracle/multilinear.rs
+++ b/crates/core/src/oracle/multilinear.rs
@@ -44,6 +44,24 @@ impl<F: TowerField> MultilinearOracleSetAddition<'_, F> {
 		Ok(self.mut_ref.add_to_set(oracle))
 	}
 
+	pub fn structured(self, n_vars: usize, expr: ArithCircuit<F>) -> Result<OracleId, Error> {
+		if expr.binary_tower_level() > F::TOWER_LEVEL {
+			bail!(Error::TowerLevelTooHigh {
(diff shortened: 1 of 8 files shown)


crates/core/src/oracle/multilinear.rs
```diff
@@ -601,15 +620,16 @@ impl DeserializeBytes for MultilinearPolyVariant<BinaryField128b> {
 		Ok(match u8::deserialize(&mut buf, mode)? {
 			0 => Self::Committed,
 			1 => Self::Transparent(DeserializeBytes::deserialize(buf, mode)?),
-			2 => Self::Repeating {
+			2 => Self::Transparent2(DeserializeBytes::deserialize(buf, mode)?),
```
jimpo, line 623: Please rename to `Structured`

pepyakin: Dang, I really hoped to call it Transparent2!

Will do.

jimpo (approved):

PR #472, opened 2025-05-27
[keccak] Apply selector to packed state column

jimpo:
Applying the selector to the packed state column results in fewer
constraint evaluations.


HiddenAndBound (approved):
LGTM

PR #480, opened 2025-05-27
[core] introduce SizedConstraintSet

pepyakin:
As part of
[CRY-363: Table-Aware Constraint System](https://linear.app/irreducible/issue/CRY-363/table-aware-constraint-system)
we want to make the constraint system description into the transcript. What
prevents us from doing so is that the sizes of a particular instance are
embedded in the constraint system. Besides MultilinearOracleSet the sizes
are present in the ConstraintSet. The changeset presented here aims to
reduce the scope of n_vars usage just to the proving/verifying, where the
newly introduced SizedConstraintSet is used now.

Note that we still have the n_vars in the ConstraintSet, but removing it should
be pretty straightforward in a follow-up PR that will actually switch over to
passing the table sizes through the transcript.

### DIFF
--- a/crates/core/src/constraint_system/prove.rs
+++ b/crates/core/src/constraint_system/prove.rs
@@ -40,7 +40,7 @@ use crate::{
 	},
 	fiat_shamir::{CanSample, Challenger},
 	merkle_tree::BinaryMerkleTreeProver,
-	oracle::{Constraint, MultilinearOracleSet, OracleId},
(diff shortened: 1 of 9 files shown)


pepyakin:
In v4 of this PR, I removed the `#[deprecated]` marker, because I did not want to deal with this non-sense, esp. considering that it is going away soon.

jimpo (approved):

PR #491, opened 2025-05-28
SYS-258: fix `test_generic_fri_fold` host buffer

BenjaminTrapani:
The prior version invoked `copy_h2d` with a host buffer that was not allocated with the host allocator. Using a buffer allocated with the default allocator results in incorrect behavior if the hardware implementation derives additional addressing information during copies from the provided slice assuming that the slice was allocated by the implementation's host allocator.


GraDKh (approved):
Probably we need to have some wrapper over a memory slice and return it from host allocator. And use it in `copy_h2d` and `copy_d2h` function. So that we are sure that all the host memory is allocated via the proper allocator

PR #503, opened 2025-05-29
Replace MIN_SLICE_LEN with ALIGNMENT in the device memory

GraDKh:
This main purpose of this change to support arbitrary small slices of the "device" memory in `FastCpuComputeLayer`.

The main idea is to represent `FSlice`/`FSliceMut` as an enum with two options:
 - Borrowed slice, when the number of elements is a multiple of `P::WIDTH`
 - Owned storage with a single packed element for the case when the length is strictly less than `P::WIDTH`.

We still can't represent slices bigger than `P::WIDTH` which aren't aligned by the `P::WIDTH` border but that seems to be sufficient for our needs, e.g. we are always able to split pow of two length slices by half.

The changes introduced by this PR:
 - replace `ComputeMemory::MIN_SLICE_LEN` with `ComputeMemory::ALIGNMENT`
 - relax conditions for all the slice spliting method, now the slices with the non-aligned length are allowed when their length is strictly less than the alignment
 - change the implementation of `FSlice` and `FSliceMut` in the `binius_fast_compute` crate.

The HW and `CPULayer` implementations of the memory do not need to change anything apart from applying the renaming.

### DIFF
--- a/crates/compute/src/memory.rs
+++ b/crates/compute/src/memory.rs
@@ -67,7 +67,7 @@ impl<Slice: SizedSlice> SlicesBatch<Slice> {
 
 /// Interface for manipulating handles to memory in a compute device.
 pub trait ComputeMemory<F> {
-	const MIN_SLICE_LEN: usize;
+	const ALIGNMENT: usize;
 
 	/// An opaque handle to an immutable slice of elements stored in a compute memory.
 	type FSlice<'a>: Copy + SizedSlice + Sync;
@@ -93,14 +93,14 @@ pub trait ComputeMemory<F> {
 	///
 	/// ## Preconditions
 	///
-	/// - the range bounds must be multiples of [`Self::MIN_SLICE_LEN`]
+	/// - the range bounds must be multiples of [`Self::ALIGNMENT`]
 	fn slice(data: Self::FSlice<'_>, range: impl RangeBounds<usize>) -> Self::FSlice<'_>;
 
 	/// Borrows a subslice of a mutable memory slice.
 	///
 	/// ## Preconditions
 	///
-	/// - the range bounds must be multiples of [`Self::MIN_SLICE_LEN`]
+	/// - the range bounds must be multiples of [`Self::ALIGNMENT`]
 	fn slice_mut<'a>(
 		data: &'a mut Self::FSliceMut<'_>,
 		range: impl RangeBounds<usize>,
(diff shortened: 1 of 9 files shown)


crates/compute/src/memory.rs
```diff
@@ -89,18 +95,20 @@ pub trait ComputeMemory<F> {
 	/// This allows the immutable reference to be copied.
 	fn as_const<'a>(data: &'a Self::FSliceMut<'_>) -> Self::FSlice<'a>;
 
-	/// Borrows a subslice of an immutable memory slice.
+	/// Borrows a subslice of an immutable memory slice
 	///
 	/// ## Preconditions
 	///
-	/// - the range bounds must be multiples of [`Self::MIN_SLICE_LEN`]
+	/// - the resulting slice must be either aligned by [`Self::ALIGNMENT`] or have a length
```
jimpo, line 102: This precondition is weird. I'd rather take the approach where `ALIGNMENT` is required to be power-of-two and we have `split_half` & `split_half_mut`.

GraDKh: That would simplify the implementation a bit, but as an interface that methods looks like an ad-hoc solution - a special splitting logic for the special case. Moreover I'm not sure if it covers all the cases, probably when splitting into equal chunks we may all face the case when a single chunks is smaller than packing width.

To me the interface where all the splitting functions follow the same rules looks more consistent. What do you think?

jimpo: I'm pretty confident it will cover all of our needs. The buffers we operate on are all power-of-two sized, and when we split them, we divide into a power-of-two number of sub-blocks. The main exception is the allocator, which could work around this if a small buffer is ever requested.

crates/compute/src/memory.rs
```diff
@@ -67,7 +67,7 @@ impl<Slice: SizedSlice> SlicesBatch<Slice> {
 
 /// Interface for manipulating handles to memory in a compute device.
 pub trait ComputeMemory<F> {
-	const MIN_SLICE_LEN: usize;
+	const ALIGNMENT: usize;
```
jimpo, line 70:
```suggestion
	/// The required alignment of indices for the split methods. This must be a power of two.
	const ALIGNMENT: usize;
```

jimpo (dismissed):

PR #512, opened 2025-05-30
[core] Replace RegularSumcheckProver with BivaraiteSumcheckProver in piop::prove

jkirmayer:
### TL;DR

Refactored the PIOP proving system to use the new compute layer abstraction, enabling GPU acceleration for sumcheck proofs.

### What changed?

- Added `Debug` trait bounds to `ComputeMemory` and its associated types
- Modified `prove` function in `piop/prove.rs` to use the new `ComputeLayer` abstraction instead of `ComputationBackend`
- Replaced the `RegularSumcheckProver` with `BivariateSumcheckProver` for better performance
- Updated the constraint system's `prove` function to use `CpuLayer` with explicit memory allocation
- Simplified test code to use the new compute layer approach

### How to test?

Run the existing test suite, particularly focusing on:
- `test_with_one_poly`
- `test_without_opening_claims`
- `test_with_one_n_vars`
- `test_commit_prove_verify_extreme_rate`
- `test_commit_prove_verify_small`
- `test_commit_prove_verify`

### Why make this change?

This change is part of a larger effort to abstract computation backends, allowing the proving system to leverage hardware acceleration. By using the new `ComputeLayer` abstraction, the code can now efficiently allocate and manage memory across different devices (CPU/HAL), which should significantly improve performance for sumcheck proofs. The change also simplifies the API by making memory management more explicit and predictable.


crates/core/src/piop/prove.rs
```diff
@@ -231,34 +221,69 @@ where
 		})
 		.collect::<Result<Vec<_>, _>>()?;
 
+	let packed_committed_fslices_mut = packed_committed_multilins
+		.iter()
+		.map(|packed_committed_multilin| {
+			let hypercube_evals = packed_committed_multilin.packed_evals().unwrap();
+			let unpacked_hypercube_evals = P::unpack_scalars(hypercube_evals);
+			let mut allocated_mem = dev_alloc.alloc(unpacked_hypercube_evals.len()).unwrap();
```
djadjka, line 229: It should be possible to add alloc::Error to the piop::Error

crates/field/src/arch/arch_optimal.rs
```diff
@@ -43,7 +43,7 @@ cfg_if! {
 		pub type OptimalUnderlier128b = crate::arch::x86_64::m128::M128;
 		pub type OptimalUnderlier256b = crate::arch::x86_64::m256::M256;
 		pub type OptimalUnderlier512b = crate::arch::x86_64::m512::M512;
-		pub type OptimalUnderlier = OptimalUnderlier512b;
+		pub type OptimalUnderlier = OptimalUnderlier128b;
```
djadjka, line 46: I'm not sure that you need to touch it.
You created the HAL with `Tower::B128 `instead of `PackedType<U, Tower::B128>`,
so you're not using PackedField — which is actually good for this case.

crates/core/src/piop/prove.rs
```diff
@@ -231,34 +221,77 @@ where
...
+	let packed_committed_fslices_mut = packed_committed_multilins
+		.iter()
+		.map(|packed_committed_multilin| {
+			let hypercube_evals = packed_committed_multilin
+				.packed_evals()
+				.expect("Prover should always populate witnesses");
+			let unpacked_hypercube_evals = P::unpack_scalars(hypercube_evals);
+			let mut allocated_mem = dev_alloc
+				.alloc(1 << packed_committed_multilin.n_vars())
+				.map_err(Error::Alloc)?;
```
djadjka, line 233: You don't need .map_err(Error::Alloc) here — the ? operator will call Error::from() automatically.

djadjka:
* **#562** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#512** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add the label <kbd>merge-ready</kbd> to this PR to add it to the merge queue.

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/IrreducibleOSS?ref=merge-queue-instructions-in-stack-comment&prId=512).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

crates/core/src/constraint_system/prove.rs
```diff
@@ -476,19 +477,61 @@ where
...
+		"PIOP HAL Setup",
+		phase = "piop_compiler",
+		perfetto_category = "phase.sub"
+	)
+	.entered();
+
+	let hal = FastCpuLayer::<Tower, PackedType<U, Tower::B128>>::default();
+
+	let host_mem_size_committed = committed_multilins.len();
+	let dev_mem_size_committed = committed_multilins
```
djadjka, line 491: I'm not sure that this is correct, because you're copying `packed_evals()` to device memory — so the difference will be multiplied by the extension degree.

crates/compute/src/memory.rs
```diff
@@ -71,10 +71,10 @@ pub trait ComputeMemory<F> {
 	const ALIGNMENT: usize;
 
 	/// An opaque handle to an immutable slice of elements stored in a compute memory.
-	type FSlice<'a>: Copy + SizedSlice + Sync;
+	type FSlice<'a>: Copy + SizedSlice + Sync + Debug;
```
djadjka, line 74: I'm not sure we need to derive Debug here, since it's unclear what that would mean for non-CPU implementations.
Maybe @jimpo, @GraDKh, or @BenjaminTrapani have an answer.

crates/core/src/constraint_system/prove.rs
jkirmayer: These are all large-field sumchecks, so the embedding degree is just 1, unless I'm missing something massive. (BivariateProductSumcheck just uses hal's extrapolate line on a large-field FSlice)

crates/core/src/protocols/sumcheck/v3/bivariate_product.rs
```diff
@@ -207,10 +216,11 @@ where
 
 		// Copy the fully folded multilinear evaluations to the host.
 		let buffer = self.host_alloc.alloc(self.multilins.len())?;
-		for (multilin, dst_i) in iter::zip(self.multilins, &mut *buffer) {
-			let vals = multilin.const_slice();
+
+		for i in 0..self.multilins.len() {
```
jimpo, line 220: What does this do differently now?

jkirmayer: Nothing, just rewrote for my own readability, will switch back now

crates/core/src/constraint_system/prove.rs
jkirmayer: Packed committed polys are embedded into the larger field already before the sumcheck takes place. They're over LOG_DEGREE fewer n_vars as a result, but that all gets handled before this point

crates/core/src/protocols/sumcheck/v3/bivariate_product.rs
djadjka: I think it steal not switched back

djadjka (commented):
Left a few comments. 
Overall it looks good to me, but I’d feel more confident if @jimpo reviewed it too before merging.

crates/core/src/ring_switch/tests.rs
```diff
@@ -310,26 +312,40 @@ fn commit_prove_verify_piop<U, F, MTScheme, MTProver>(
...
 
-	let backend = make_portable_backend();
 	let ReducedWitness {
 		transparents: transparent_multilins,
 		sumcheck_claims,
 	} = prove(&system, &committed_multilins, &mut proof, MemoizedData::new()).unwrap();
 
-	let domain_factory = DefaultEvaluationDomainFactory::<B8>::default();
+	let hal = CpuLayer::<Tower>::default();
+	let host_mem_size_committed = committed_multilins.len();
```
jimpo, line 321: This is precise, but not necessary. It's just a test -- you can just hardcode a value that's large enough to work and keep rerunning it until it does. Allocating 64 KiB - 1 MiB should be totally fine.

crates/core/src/protocols/sumcheck/v3/bivariate_product.rs
```diff
@@ -119,6 +118,7 @@ where
 			.iter()
 			.map(|multilin| multilin.const_slice())
 			.collect::<Vec<_>>();
+
```
jimpo, line 121: There are a lot of these random newline corrections in this diff. Please try to avoid that.

PR #570, opened 2025-06-04
[compute] Create KernelBuilder trait

jimpo:
Move kernel construction methods from ComputeLayer onto a new trait, which is an associated type. This helps move towards an API where ComputeLayer can be sync while allowing for multithreaded backends.

### DIFF
--- a/crates/compute/src/layer.rs
+++ b/crates/compute/src/layer.rs
@@ -24,7 +24,7 @@ pub trait ComputeLayer<F: Field>: 'static + Sync {
 
 	/// The executor that can execute operations on a kernel-level granularity (i.e., a single
 	/// core).
-	type KernelExec: KernelBuilder<F, Mem = Self::DevMem, Value = Self::KernelValue, ExprEval = Self::ExprEval>;
+	type KernelExec: KernelExecutor<F, Mem = Self::DevMem, Value = Self::KernelValue, ExprEval = Self::ExprEval>;
 
 	/// The operation (scalar) value type.
 	type OpValue;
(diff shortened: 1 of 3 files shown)


crates/compute/src/layer.rs
```diff
@@ -448,6 +454,72 @@ pub trait ComputeLayer<F: Field>: 'static + Sync {
 	) -> Result<(), Error>;
 }
 
+/// An interface for constructing execution kernels.
+///
+/// A _kernel_ is a program that executes synchronously in one thread, with access to
+/// local memory buffers.
+///
+/// See [`ComputeLayer::accumulate_kernels`] for more information.
+pub trait KernelBuilder<F> {
```
GraDKh, line 463: I'm only confused about the terminology consistency. As for be it's pretty weird to see a builder, that doesn't have any build methods

GraDKh (dismissed):

crates/compute/src/layer.rs
jimpo: Yeah, that's fair. I'm ambivalent about Executor vs Builder as far as terminology. In the IC1 context, builder makes sense because it's building a description of a kernel program, which will then be compiled and executed. In the CPU context, it executes immediately.

@BenjaminTrapani I assume you're OK if we name the trait KernelExecutor?

BenjaminTrapani: Renaming to `KernelExecutor` sounds good to me. This component is called `KernelBuilder` in ic1 because it defines the kernel layout and compiler input for a transformation corresponding to the chunk size. The built artifacts consist of the compiled instructions, input / output layouts and kernel-local memory layout.

GraDKh (approved):

crates/compute/src/layer.rs
```diff
@@ -448,6 +454,72 @@ pub trait ComputeLayer<F: Field>: 'static + Sync {
...
+	type Mem: ComputeMemory<F>;
+
+	/// The kernel(core)-level operation (scalar) type. This is a promise for a returned value.
+	type Value;
+
+	/// The evaluator for arithmetic expressions (polynomials).
+	type ExprEval: Sync;
+
+	/// Declares a kernel-level value.
+	fn decl_value(&mut self, init: F) -> Result<Self::Value, Error>;
```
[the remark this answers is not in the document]
jimpo: I don't understand the question exactly, but this sounds like a follow-up issue to me.

PR #590, opened 2025-06-04
[core] track maxrss

pepyakin:


pepyakin:
* **#590** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius/590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add the label <kbd>merge-ready</kbd> to this PR to add it to the merge queue.

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/IrreducibleOSS?ref=merge-queue-instructions-in-stack-comment&prId=590).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

GraDKh (commented):
TBH If we want to measure this I would better use perfetto functionality rather then doing it manually:
https://perfetto.dev/docs/data-sources/memory-counters

djadjka:
> So I gotta ask the question: what would be the benefits of using the built-in memory counters and is it worth it to not include this, even if there is intention of improving it in the future?

Maybe I'm wrong, but I think this approach adds noise, since it probably includes Perfetto's own memory.
That’s fine for large witnesses, but could be misleading when analyzing smaller ones.

Also, we have aggregation of counter, which can be confusing if we want to measure memory usage more deeply within specific parts of the code.

Maybe @jimpo knows whether it's actually useful for us — and if it is, perhaps @fkondej could look into how hard it would be to integrate perfetto functionality into our codebase.

pepyakin:
> Maybe I'm wrong, but I think this approach adds noise, since it probably includes Perfetto's own memory.

Looking at the linked article, those data sources are using `/proc`, so should be equivalent in that regard.

FWIW, I am not married to this PR either and not the hill I am dying on, so feel free to close it if you feel strongly.

GraDKh:
>Maybe I'm wrong, but I think this approach adds noise, since it probably includes Perfetto's own memory.

jimpo:
Agree with @pepyakin that this is a cheap addition that can add some useful insight. Even if it's not perfect, we should merge it until we have something better. *But*, it needs to be documented with a brief mentions of any limitations from this convo.

crates/core/src/constraint_system/prove.rs
```diff
@@ -511,6 +522,32 @@ where
 	Ok(proof)
 }
 
+fn emit_max_rss() {
```
jimpo, line 525: Please move to a util module. `binius_utils::perf` or something. Also, this needs a docstring.

Regarding naming, it seems like it should be `log_max_rss` instead of `edit`, since it's logging a tracing event?

pepyakin: It's actually **e**mit, as in emitting a tracing event, but I don't care about the name that much.

Now, regarding moving to the utils module. That would add `tracing` dep to the utils module. Are we alright with that?

jimpo: Oops, I misread. Naming is fine.

Regarding location, I'm happy with either a `binius_core::util` module or a `tracing` Cargo feature in the `binius_utils` crate.

GraDKh:
I've made the perfetto solution working... And agree that it's a no-go, as it requires ftrace (which is not present on MacOS) , perfetto binaries available, and launching a process in a privileged mode to capture system-wide events. It's super inconvinient

crates/core/src/constraint_system/prove.rs
GraDKh: We can actually place it in `tracing-profile` as this may be handy in `ic1` as well

PR #594, opened 2025-06-04
[compute] Relax requirement on kernel memory

jimpo:
Kernel memory buffers don't need to be the same type as global memory
buffers.


crates/compute/src/layer.rs
```diff
@@ -599,6 +599,10 @@ pub type FSlice<'a, F, HAL> = <<HAL as ComputeLayer<F>>::DevMem as ComputeMemory
 pub type FSliceMut<'a, F, HAL> =
 	<<HAL as ComputeLayer<F>>::DevMem as ComputeMemory<F>>::FSliceMut<'a>;
 
+pub type KernelMem<F, HAL> = <<HAL as ComputeLayer<F>>::KernelExec as KernelExecutor<F>>::Mem;
```
BenjaminTrapani, line 602: Do any of the HAL's kernel-specific functions need to be updated to accept these new types as inputs?

BenjaminTrapani (approved):

PR #602, opened 2025-06-04
[bench] migrate b32_mul to m3

pepyakin:
See [CRY-432: b32_mul](https://linear.app/irreducible/issue/CRY-432/b32-mul)

### DIFF
--- a/examples/b32_mul.rs
+++ b/examples/b32_mul.rs
@@ -1,18 +1,20 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-// Uses binius_circuits which is being phased out.
-#![allow(deprecated)]
-
 use anyhow::Result;
-use binius_circuits::builder::{ConstraintSystemBuilder, types::U};
 use binius_core::{constraint_system, fiat_shamir::HasherChallenger};
-use binius_field::{BinaryField32b, TowerField, tower::CanonicalTowerFamily};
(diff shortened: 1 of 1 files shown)


examples/b32_mul.rs
```diff
@@ -38,83 +38,74 @@ fn main() -> Result<()> {
...
-		>(&constraint_system, args.log_inv_rate as usize, SECURITY_BITS, &[], witness, &backend)?;
+		>(
+			&ccs, args.log_inv_rate as usize, SECURITY_BITS, &[], witness, &make_portable_backend()
+		)?;
 
 	println!("Proof size: {}", ByteSize::b(proof.get_proof_size() as u64));
 
 	constraint_system::verify::<
-		U,
+		OptimalUnderlier128b,
```
djadjka, line 103: you should use OptimalUnderlier instead of OptimalUnderlier128b

pepyakin: Thanks! I was going back and forth with it. I thought I had a problem on unit-test-amd when I got an error. I don't remember exactly but it was something about that `M512`, so I just used this one. 

I will revert it back here and in other places. Although I might need to do that as one pass as a follow up  because there are others (and some even mix). Is that alright with you?

HiddenAndBound (commented):
LGTM

PR #609, opened 2025-06-05
[examples] [hash] Parallelized architecture-specific hashing used in examples

jkirmayer:
### TL;DR

Updated the constraint system to use parallel Groestl256 hashing for improved performance.

### What changed?

- Modified the constraint system's `prove` function to use `ParallelDigest` trait instead of the `Digest` trait for its hasher
- Updated the `Groestl256Multi` implementation to handle empty data arrays by filling them with zeroes
- Changed all examples to use `Groestl256Parallel` instead of `Groestl256` for hashing

### How to test?

Run any of the example programs to verify they still work correctly with the parallel hashing implementation. The examples should execute faster due to the parallel hashing optimization.

### Why make this change?

This change improves performance by using parallel hashing capabilities. The parallel Groestl256 implementation can process multiple data streams simultaneously, which speeds up the proving process. The modification to handle empty data arrays also makes the implementation more robust by allowing it to process inputs of varying lengths.


jimpo:
@jkirmayer Is this on Graphite? I don't see it there

jimpo (approved):

PR #611, opened 2025-06-05
[bench] migrate bitwise_ops to m3

pepyakin:
See [CRY-438: bitwise](https://linear.app/irreducible/issue/CRY-438/bitwise)

### DIFF
--- a/examples/bitwise_ops.rs
+++ b/examples/bitwise_ops.rs
@@ -1,19 +1,23 @@
 // Copyright 2024-2025 Irreducible Inc.
 
-// Uses binius_circuits which is being phased out.
-#![allow(deprecated)]
-
 use std::{fmt::Display, str::FromStr};
 
 use anyhow::Result;
-use binius_circuits::builder::{ConstraintSystemBuilder, types::U};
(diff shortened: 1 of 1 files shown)


examples/bitwise_ops.rs
```diff
@@ -52,14 +58,118 @@ struct Args {
...
+	Or(BitwiseOr),
+}
+
+impl BitwiseGadget {
+	fn populate<P>(&self, segment: &mut TableWitnessSegment<P>) -> Result<()>
+	where
+		P: binius_field::PackedFieldIndexable<Scalar = B128> + binius_field::PackedExtension<B1>,
+	{
+		match *self {
+			BitwiseGadget::And(ref gadget) => gadget.populate(segment),
```
HiddenAndBound, line 166: nit: why not use &?

HiddenAndBound (approved):
LGTM.

examples/bitwise_ops.rs
pepyakin: Actually a good point. We don't need to dereference *self here and that would make `gadget` a shared reference automatically.

HiddenAndBound (approved):
LGTM

PR #617, opened 2025-06-05
SYS-287: move HAL operations to exec

BenjaminTrapani:
Remove the `Sync` requirement from the HAL. This property was only required by the optimized CPU implementation.

### DIFF
--- a/crates/compute/src/layer.rs
+++ b/crates/compute/src/layer.rs
@@ -15,22 +15,12 @@ use super::{
 use crate::memory::{SizedSlice, SlicesBatch};
 
 /// A hardware abstraction layer (HAL) for compute operations.
-pub trait ComputeLayer<F: Field>: 'static + Sync {
+pub trait ComputeLayer<F: Field>: 'static {
 	/// The device memory.
 	type DevMem: ComputeMemory<F>;
 
 	/// The executor that can execute operations on the device.
-	type Exec;
-
-	/// The executor that can execute operations on a kernel-level granularity (i.e., a single
-	/// core).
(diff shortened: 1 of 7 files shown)


crates/compute/src/layer.rs
```diff
@@ -15,22 +15,12 @@ use super::{
...
-	/// The executor that can execute operations on a kernel-level granularity (i.e., a single
-	/// core).
-	type KernelExec: KernelExecutor<F, ExprEval = Self::ExprEval>;
-
-	/// The operation (scalar) value type.
-	type OpValue;
-
-	/// The evaluator for arithmetic expressions (polynomials).
-	type ExprEval: Sync;
+	type Exec<'a>: ComputeLayerExecutor<F, DevMem = Self::DevMem>;
```
jimpo, line 23: Why does this have an 'a lifetime? That appears to be unused.

jimpo:
Great refactor, very exciting!

crates/compute/src/layer.rs
jimpo: Yep, got it, thanks for the explanation.

I'm actually not sure this API could support executors referencing local data created in `execute` and that's a limitation of the closure pattern. We don't need to tackle it now though, the solution will be to eventually remove `execute` altogether and put a `finish(self)` method or something on `Executor`.

Reminds me of this refactor I had to do in the old linerate repo https://github.com/IrreducibleOSS/linerate/commit/b3488cc698394e187cf3d94706185b42f4b61358

jimpo (approved):

PR #635, opened 2025-06-06
[m3] Add read method for lookup tables

HiddenAndBound:
# Add read and write methods to TableBuilder

This PR adds two new methods to `TableBuilder`:

1. `read<FSub, const V: usize>` - Reads a group of columns from a specified lookup table, enforcing that values are obtained through the specified lookup channel.

A test case for the `read` method is included to verify functionality with stacked columns.

### DIFF
--- a/crates/m3/src/builder/table.rs
+++ b/crates/m3/src/builder/table.rs
@@ -657,6 +657,49 @@ impl<'a, F: TowerField> TableBuilder<'a, F> {
 		});
 	}
 
+	/// Writes a group of columns from a specified lookup table.
+	///
+	/// This method enforces that the values of the provided columns are obtained from a lookup
+	/// table through the specified lookup channel. The lookup channel identifies the target
+	/// table for writing values.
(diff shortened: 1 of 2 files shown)


jimpo (approved):

crates/m3/src/builder/table.rs
```diff
@@ -839,4 +888,40 @@ mod tests {
...
+			FlushOpts {
+				multiplicity: 1 << LOG_STACKING_FACTOR as u32,
+				selectors: vec![],
+			},
+		);
+
+		let alloc = Bump::new();
+		let mut witness: WitnessIndex<PackedType<OptimalUnderlier, B128>> =
+			WitnessIndex::new(&cs, &alloc);
+		witness.fill_constant_cols().unwrap();
```
pepyakin, line 920: Why is it here and does it do anything? I am asking because it seems that this should not do anything with tables that haven't been initialized before.

PR #671, opened 2025-06-08
HAL operation for map_kernels

djadjka:


crates/compute/src/layer.rs
```diff
@@ -181,6 +181,59 @@ pub trait ComputeLayerExecutor<F: Field> {
...
+	///   unchanged.
+	///
+	/// [`ComputeLayer`] implementations are free to call the specification closure multiple times,
+	/// for example with different values for `log_chunks`.
+	///
+	/// ## Arguments
+	///
+	/// * `map` - the kernel specification closure. See the "Kernel specification" section above.
+	/// * `mem_maps` - the memory mappings for the kernel-local buffers.
+	fn map_kernels(
```
GraDKh, line 226: We need to add some tests using this functionality

crates/compute/src/cpu/layer.rs
```diff
@@ -119,6 +119,33 @@ impl<F: TowerTop> CpuLayerExecutor<F> {
...
+	fn process_kernels_chunks<R>(
+		&self,
+		map: impl Sync
+		+ for<'a> Fn(
+			&'a mut CpuKernelBuilder,
+			usize,
+			Vec<KernelBuffer<'a, F, CpuMemory>>,
+		) -> Result<R, Error>,
+		mut mem_maps: Vec<KernelMemMap<'_, F, CpuMemory>>,
+	) -> Result<Vec<R>, Error> {
```
GraDKh, line 132: It's a pity you that your're allocating this vector and actually don't use it in both cases.
Maybe better return `impl Iterator<Item = Result<R, Error>>` ?

crates/compute/src/cpu/layer.rs
```diff
@@ -119,6 +119,33 @@ impl<F: TowerTop> CpuLayerExecutor<F> {
...
+		(0..1 << log_chunks)
+			.map(|i| {
+				let local_buffer_alloc = BumpAllocator::new(local_buffer.as_mut());
+				let kernel_data =
+					Self::map_kernel_mem(&mut mem_maps, &local_buffer_alloc, log_chunks, i);
+				map(&mut CpuKernelBuilder, log_chunks, kernel_data)
+			})
+			.collect::<Result<Vec<_>, Error>>()
+	}
 }
```
GraDKh, line 149: Please not that this code has changed in main, so merge it with care

GraDKh (approved):

PR #676, opened 2025-06-08
[compute] Move eq_ind_partial_eval helper function to library

jimpo:
This moves a common function from a test helper method to the binius_compute crate library.

Renamed function to reduce confusion with ComputeLayerExecutor::tensor_expand and to better match the function in `binius_math` crate that does the same functionality without the ComputeLayer.

### DIFF
--- a/crates/compute/src/lib.rs
+++ b/crates/compute/src/lib.rs
@@ -10,6 +10,7 @@ pub mod alloc;
 pub mod cpu;
 pub mod layer;
 pub mod memory;
+pub mod ops;
 
 pub use layer::*;
 pub use memory::*;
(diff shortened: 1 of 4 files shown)


djadjka:
Maybe you’d prefer to use https://github.com/IrreducibleOSS/binius/blob/c82652b3843dd8c719bd66cf998e6e160a3a85ec/crates/core/src/compute_math/tensor_prod_eq_ind.rs

jimpo:
Oh, yes, thanks.

djadjka (approved):
Since Send + Sync bounds were removed from Hal, the code I previously shared is now outdated.
This PR is good to merge as-is.

PR #694, opened 2025-06-09
[core] fix piop commit metadata indexing

pepyakin:
Prior to this change, `make_oracle_commit_meta` makes an assumption that oracles
IDs are just gapless sequence of indicies. While this is true at the moment,
this won't be true in presence of [CRY-363: Table-Aware Constraint System](https://linear.app/irreducible/issue/CRY-363/table-aware-constraint-system)
, because some of the tables are going to be 0-sized and their oracles will
not be instantiated.


HiddenAndBound (approved):
LGTM

PR #696, opened 2025-06-09
[m3] Use lookup tables for Keccak's bitwise AND operation

HiddenAndBound:
# keccak with lookups for AND

This PR implements bitwise AND operations in the Keccak permutation using lookup tables instead of direct constraints. The implementation:

- Adds a lookup channel to the Keccak gadget
- Creates packed columns for bitwise operations
- Implements merge functions for AND operations
- Updates the chi step to use lookup-based AND operations
- Modifies the test cases and examples to support the new lookup approach

### DIFF
--- a/crates/m3/src/gadgets/hash/keccak/lookedup.rs
+++ b/crates/m3/src/gadgets/hash/keccak/lookedup.rs
@@ -614,7 +614,7 @@ impl std::ops::Index<usize> for PerBatchLens<'_> {
 mod tests {
 	use std::cmp::Reverse;
 
-	use binius_field::{arch::OptimalUnderlier128b, as_packed_field::PackedType};
+	use binius_field::{arch::OptimalUnderlier, as_packed_field::PackedType};
 	use bumpalo::Bump;
 	use itertools::Itertools;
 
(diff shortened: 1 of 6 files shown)


crates/m3/src/gadgets/hash/keccak/lookedup.rs
```diff
@@ -0,0 +1,759 @@
...
+				index.get_mut_as(self.b1_and_b2[(x, y)])?;
+			for i in 0..b2.len() {
+				// B[x,y] xor ((not B[x+1,y]) and B[x+2,y])
+				let in_a = b1[i].val();
+				let in_b = b2[i].val();
+				let output = in_a & in_b;
+				b1_and_b2[i] = output.into();
+				merged[i] = merge_bitand_vals(in_a, in_b, output).into();
+			}
+		}
```
pepyakin, line 489: Nit, small one: literally everywhere in this module we use `for x in 0..5 { for y in 0..5 { } }` and only here it's the inverse.

crates/m3/src/gadgets/hash/keccak/lookedup.rs
```diff
@@ -0,0 +1,759 @@
...
+			boundaries: vec![],
+			table_sizes: witness.table_sizes(),
+		};
+
+		let ccs = cs.compile(&statement).unwrap();
+		let witness = witness.into_multilinear_extension_index();
+
+		binius_core::constraint_system::validate::validate_witness(
+			&ccs,
+			&[],
```
pepyakin, line 753: nit: would be better to use `statement.boundaries`. Noticed that it's easy to trip up otherwise.

pepyakin (approved):
Looks good.

This uses quite a bit of duplication but I assume it's fine. If this is an experimental arithmetization, then it's ok, and does not require cleaning up that much. If it's here to stay, it's also alright, since it could be cleaned up as a follow up, or when we start duplicating the third time.

That being said, I would appreciate expanding the rustdoc of this module. Imagine you are a user of m3 and you want to use keccak. What would help you to decide whether to use one version or another? Also, it's possible that this documentation maybe is better placed in the parent module, and the stacked and lookedup modules would just refer to it.

PR #698, opened 2025-06-10
[ci] Benchmarks trigger trace processor

fkondej:


milosbackonja (approved):
Would it be more cleaner if we use python script instead inline?
It looks good to me.

PR #699, opened 2025-06-10
SYS-292: remove `host_alloc` from `ComputeLayer`

BenjaminTrapani:
The host and device allocators will be provided by the implementation-specific HAL setup logic. Binius code requiring host and device allocations should be implemented in terms of the `ComputeAllocator` interface. Update existing callsites performing host allocations using the HAL API to accept a `ComputeAllocator` returning host memory.


crates/compute/src/cpu/alloc.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub struct CpuComputeAllocator<F> {
+	data: Vec<F>,
+}
+
+impl<F> CpuComputeAllocator<F>
+where
+	F: Sync + Debug + Send + 'static,
+{
+	pub fn into_inner(&mut self) -> HostBumpAllocator<'_, F> {
```
GraDKh, line 17: The naming seems a bit weird to me, don't quite understand why "inner", usually "inner" means that some object is stored inside. Maybe `to_bump_allocator`?

GraDKh (approved):

PR #709, opened 2025-06-10
[core] simplify constraint set builder

pepyakin:
This functionality is rudimentary and was needed for when the constraint sets
were built in binius_circuits. Now, we don't need that but it really is
annoying to maintain this. So — remove.

### DIFF
--- a/crates/core/src/oracle/constraint.rs
+++ b/crates/core/src/oracle/constraint.rs
@@ -7,9 +7,8 @@ use binius_field::{Field, TowerField};
 use binius_macros::{DeserializeBytes, SerializeBytes};
 use binius_math::{ArithCircuit, CompositionPoly};
 use binius_utils::bail;
-use itertools::Itertools;
 
-use super::{Error, MultilinearOracleSet, MultilinearPolyVariant, OracleId};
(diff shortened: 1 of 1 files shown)


crates/core/src/oracle/constraint.rs
```diff
@@ -105,20 +104,6 @@ impl<F: Field> ConstraintSetBuilder<F> {
...
-		&mut self,
-		name: impl ToString,
-		oracle_ids: impl IntoIterator<Item = OracleId>,
-		composition: ArithCircuit<F>,
-	) {
-		self.constraints.push(UngroupedConstraint {
-			name: name.to_string(),
-			oracle_ids: oracle_ids.into_iter().collect(),
-			composition,
-			predicate: ConstraintPredicate::Zero,
```
djadjka, line 118: If you plan to remove `add_zerocheck`, you will also need to refactor `ConstraintPredicate` and all usages of it.

pepyakin: oh yeah I noticed that it's not used anywhere. It even panics in validate_witness it. You wan't me to do this in this PR?

djadjka (approved):
LGTM

PR #727, opened 2025-06-11
Introduce ComputeHolder and use it in generic HAL tests

GraDKh:


BenjaminTrapani (approved):

crates/compute/src/layer.rs
```diff
@@ -604,6 +608,20 @@ pub type KernelSlice<'a, 'b, F, HAL> = <KernelMem<'b, F, HAL> as ComputeMemory<F
 pub type KernelSliceMut<'a, 'b, F, HAL> =
 	<KernelMem<'b, F, HAL> as ComputeMemory<F>>::FSliceMut<'a>;
 
+/// This is a trait for a holder type for the popular triple:
+/// * a compute layer (HAL),
+/// * a host memory allocator,
+/// * a device memory allocator.
+pub trait ComputeHolder<F: Field, HAL: ComputeLayer<F>> {
+	fn to_data(&mut self) -> ComputeData<'_, F, HAL>;
```
BenjaminTrapani, line 616: What do you think about having `ComputeData` be a trait that requires `host_alloc` and `dev_alloc` to implement `ComputeAllocator<'a, F, CpuMemory` and `ComputeAllocator<'a, F, HAL::DevMem>` respectively? This would let implementations define custom fast bump allocators that place lower bounds on the size of the allocated memory than could be placed if large slices of host and device memory were obtained on startup.

iradicek (approved):

crates/compute/src/layer.rs
iradicek: It sounds like a good idea, but does it need to be a trait?

PR #737, opened 2025-06-12
Optimize `FastCpuLayer::accumulate_kernels` for the case when there are many chunks

GraDKh:
### TL;DR

Optimize the `FastCpuLayer::kernel_map` function by using `StridedArray2DViewMut`. This allows to parallelize the chunk's memory buffers initialization which takes significant amount of time when there are many memory buffers and the number of chunks is big (which is a case when many CPU cores are available).

### What changed?

- Moved `strided_array` module from `crates/ntt` to `crates/utils` to make it more widely available
- Enhanced `StridedArray2DViewMut` with a new `iter_column_mut` method for efficient column-wise iteration
- Refactored `ComputeLayerExecutor` to initialize memory chunks in parallel using `StridedArray2DViewMut`
- Replaced inefficient memory initialization pattern with a more optimized approach that leverages parallel processing

### How to test?

1. Run the test suite to ensure all functionality works correctly
2. Verify that the new memory initialization pattern in `ComputeLayerExecutor` performs as expected
3. Test the new `iter_column_mut` method with the provided test case in `strided_array.rs`

### Why make this change?

The changes improve performance when initializing large memory chunks by enabling parallel initialization. Moving the `strided_array` module to the utils crate makes this useful data structure available to more components.

### DIFF
--- a/crates/fast_compute/src/layer.rs
+++ b/crates/fast_compute/src/layer.rs
@@ -3,7 +3,7 @@
 use std::{
 	any::TypeId,
 	cell::RefCell,
-	iter::{repeat_with, zip},
+	iter::zip,
 	marker::PhantomData,
 	mem::{MaybeUninit, transmute},
 	slice,
@@ -43,6 +43,7 @@ use binius_ntt::{AdditiveNTT, fri::fold_interleaved_allocated};
 use binius_utils::{
 	checked_arithmetics::{checked_int_div, strict_log_2},
 	rayon::get_log_max_threads,
+	strided_array::StridedArray2DViewMut,
 };
 use bytemuck::{Pod, zeroed_vec};
 use itertools::izip;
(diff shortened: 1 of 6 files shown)


crates/fast_compute/src/layer.rs
```diff
@@ -260,21 +261,38 @@ impl<'a, T: TowerFamily, P: PackedTop<T>> ComputeLayerExecutor<T::B128>
...
+		// - Each element is initialized by the previous `for_each` loop.
+		unsafe {
+			memory_chunks.set_len(mem_maps_count << log_chunks);
 		}
+		let memory_chunks_view = StridedArray2DViewMut::without_stride(
+			&mut memory_chunks,
+			mem_maps_count,
+			1 << log_chunks,
+		)
+		.unwrap();
```
djadjka, line 291: unwrap()

PR #740, opened 2025-06-12
Use the POLYVAL field for the flush selector sumcheck.”

djadjka:


crates/core/src/constraint_system/verify.rs
```diff
@@ -499,3 +518,92 @@ pub fn make_flush_oracles<F: TowerField>(
 		})
 		.collect()
 }
+
+fn reduce_flush_evalcheck_cliams<Tower: TowerFamily, Challenger_>(
```
jimpo, line 522: typo: cliams -> claims.

Please use `pre-commit install` to install the typos pre-commit hook.

jimpo (approved):
Nice!

PR #764, opened 2025-06-16
INF-211: Typos ci

milosbackonja:


pepyakin:
Could you elaborate a bit on what this does exactly?

milosbackonja:
> One concern I have is basically that the spell check can get annoying in case it encounters some word that it thinks a typo but it is not. What is the situation wrt dictionaries? How would I add an exception?

@pepyakin you can add exceptions in Cargo.toml.
There are already two exceptions

PR #782, opened 2025-06-16
Introduce macros to instantiate compute layer test

GraDKh:


crates/compute/test_utils/src/layer.rs
```diff
@@ -898,3 +912,144 @@ pub fn test_map_kernels<F, Hal, ComputeHolderType>(
...
+			pub type B16 = <$tower as binius_field::tower::TowerFamily>::B16;
+			pub type B32 = <$tower as binius_field::tower::TowerFamily>::B32;
+			pub type B128 = <$tower as binius_field::tower::TowerFamily>::B128;
+		}
+
+		#[test]
+		fn test_single_tensor_expand() {
+			let n_vars = 8;
+			$crate::layer::test_generic_single_tensor_expand::<fields::B128, $hal, $compute_holder>(
+				<$compute_holder>::new(1 << (n_vars + 1), 1 << n_vars),
```
BenjaminTrapani, line 931: The ic1 compute holder is created from additional parameters containing device configuration: https://github.com/IrreducibleOSS/ic1/pull/321/files#diff-f088849e87cb8cab63941cb5b8214eda7659e83e04c035429f73276bf2a3373cR84

Would it make sense to accept an expression specifying the source of the `ComputeHolder`, inferring the type from the result of this expression?

GraDKh: It is for sure not hard to apply that but as this expression will be called in a test function you won't be able to use anything apart from static variables there... Will it be OK in this case?

BenjaminTrapani: The ic1 tests require the non-static device config. Would it be possible to keep the `<$compute_holder>::new(host_size, device_size)` invocations while passing through a variable length list of identifiers from the enclosing scope?

onesk (approved):

PR #785, opened 2025-06-16
Extract macro to instantiate ring switch tests

GraDKh:
### TL;DR

Added a macro to instantiate PIOP and ring switch tests for different compute layers.

### What changed?

- Added `instantiate_piop_tests!` macro in `piop.rs` that generates six test cases for different PIOP scenarios
- Added `instantiate_ring_switch_tests!` macro in `ring_switch.rs` that generates two test cases for ring switch functionality
- Refactored existing ring switch tests to use the new macro
- Added helper functions to support the test macros
- Implemented both CPU and FastCPU test modules in `ring_switch.rs`

### How to test?

Run the tests with:
```bash
cargo test --package binius-core
```

The tests will now run for both CPU and FastCPU compute layers, verifying that the PIOP and ring switch functionality works correctly across different implementations.

### Why make this change?

This change improves test coverage by ensuring that PIOP and ring switch functionality works consistently across different compute layer implementations. By using macros to generate the tests, we avoid code duplication while still testing the same functionality with different backends. This makes it easier to add support for new compute layers in the future, as they can simply instantiate these test macros to verify compatibility.

### DIFF
--- a/crates/core/test_utils/src/ring_switch.rs
+++ b/crates/core/test_utils/src/ring_switch.rs
@@ -2,7 +2,7 @@
 
 use std::{cmp::Ordering, iter::repeat_with};
 
-use binius_compute::{ComputeHolder, ComputeLayer};
+use binius_compute::{ComputeData, ComputeHolder, ComputeLayer};
 use binius_core::{
 	fiat_shamir::HasherChallenger,
 	merkle_tree::{MerkleTreeProver, MerkleTreeScheme},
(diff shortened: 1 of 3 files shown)


crates/core/test_utils/src/ring_switch.rs
```diff
@@ -301,3 +301,119 @@ pub fn commit_prove_verify_piop<U, F, MTScheme, MTProver, Hal, HalHolder>(
...
+			&prover_sumcheck_claims,
+			hal,
+		)
+		.unwrap();
+	});
+}
+
+#[macro_export]
+macro_rules! instantiate_ring_switch_tests {
+	($compute_layer:ty, $compute_holder:ty) => {
```
djadjka, line 385: Do we need `$compute_layer` here ? `$compute_holder` should know about it

PR #800, opened 2025-06-17
FRI config to support FRI conjecture

### LINKED ISSUE #342: [fri] Support FRI configuration using the FRI conjecture
We want the ability to optionally configure the number of FRI opening queries so that it achieves the target security level assuming the FRI conjecture (see [2022/1216](https://eprint.iacr.org/2022/1216), section 3.6).

We can add a boolean parameter to [fri::calculate_n_test_queries](https://github.com/IrreducibleOSS/binius/blob/d91fe22b0c40884dad031296c58c9830bd3a26d1/crates/core/src/protocols/fri/common.rs#L279) called `fri_conjecture: bool`. The method documentation should describe this with a link to the resource above. This parameter will have to be propagated all the way up to the `binius_core::constraint_system` module.

For code cleanliness, it would be good to wrap the `log_inv_rate`, `security_bits`, and `fri_conjecture` parameters to `binius_core::constraint_sytem` into a `Params` struct.

The conjecture should _not_ be assumed by default.

roynalnaruto:
Closes #342 

### Summary

- Added a `FRISoundnessParams` struct (includes the `fri_conjecture` boolean flag)
- Added a `conjectured_per_query_err` method to calculate the per-query error assuming the conjecture 8.4
- Assumed `c1 = c2 = 1`
- Assumed `eta = 0.005` (TODO: finetune using some test results)
- Used the `total_multilins` from `CommitMeta` as the "batch size" (or `N` in the conjecture 8.4's equation)
- Refactored the rest of the codebase to use the new struct `FRISoundnessParams`
- Added inline-documentation in appropriate places

The number of queries:

| c1 = c2 | η     | log_inv_rate | RS-Code log2(dimension) |# queries proven bounds | # queries assuming conjecture |
|--------|-------|------------|-------------------------|-------------------------|--------------------------------|
| 1           | 0.005 | 1                  |  28                                    | 242                                   | 2                                                   |
| 1           | 0.005 | 2                  |  28                                    | 143                                   | 2                                                   |

### Pending
- [x] tests (with some comparison) for the conjectured # of queries
- [x] resolve merge conflicts
- [ ] finetune value for eta (η)


benediamond:
hi @roynalnaruto, this looks impressive, and we appreciate your work, but I don't think this can be merged in its current state, without further supplementary analysis done on the cryptographic side.

first of all, 2 queries (for both $\mathcal{R} = 1$ and $2$) seems absurdly low; it feels like there must be something wrong here. the distance of the code would have to be enormous for this to make sense, and I don't think one can get that kind of distance, even assuming the most aggressive conjectures. can you write down the calculation that took you to 2?

but more importantly, I am not convinced that the conjectures can simply be "dropped in" in this way. various works have tried to make BaseFold work w/ list decoding, and it is not "for free". For one, the PCS would become something like a "list PCS". in fact, WHIR appears to contain an idea where the commitment phase is actually made interactive, with an extra out-of-domain sample—that extra sample brings the thing back into a PCS, as opposed to a list PCS. but you aren't doing that here in this PR.

in short, my personal opinion is that further analysis needs to be done for me to feel comfortable merging this. let me know your thoughts; thank you.

roynalnaruto:
@benediamond Thank you for the review.

> I am not convinced that the conjectures can simply be "dropped in" in this way

You are right. I have not done any cryptographic analysis of the conjecture and how it would work out in a scheme that is a variant of FRI. 2 queries seems absurd as well.

I am going through Binius docs to understand the details. If I have something more meaningful, or at least a concrete direction, I will get back to you and re-open the work with a thorough analysis. Until then, I will close the PR for now, as it does not have any fundamental cryptographic analysis.

Thank you for the review and some ideas :)

PR #813, opened 2025-06-18
FastCompute implementation of GPA witness layer low-to-high HAL op

djadjka:
### TL;DR

Implemented the `pairwise_product_reduce` function for the FastCpuLayer and added tests for it.

### What changed?

- Implemented the previously unimplemented `pairwise_product_reduce` function in the `ComputeLayerExecutor` trait for `FastCpuLayer`
- Added input validation to ensure the input length is a power of 2 and greater than or equal to 2
- Implemented the pairwise product reduction algorithm using parallel chunks processing
- Removed the unused `fill` method from `SmallOwnedChunk`
- Simplified the `fill_constant` implementation to use the new slice abstraction
- Added necessary imports (`Itertools` and `PackedMemorySlice`)
- Added tests for the new functionality with both single round and multi-round reductions

### How to test?

Run the new tests:
```
cargo test -p fast_compute test_pairwise_product_reduce
cargo test -p fast_compute test_pairwise_product_reduce_single_round
```

### Why make this change?

This change implements a previously unimplemented function that is needed for computing pairwise products and reducing them, which is a common operation in cryptographic protocols. The implementation is optimized for CPU execution using parallel processing where possible.

### DIFF
--- a/crates/fast_compute/src/layer.rs
+++ b/crates/fast_compute/src/layer.rs
@@ -46,12 +46,12 @@ use binius_utils::{
 	strided_array::StridedArray2DViewMut,
 };
 use bytemuck::{Pod, zeroed_vec};
-use itertools::izip;
+use itertools::{Itertools, izip};
 use thread_local::ThreadLocal;
 
 use crate::{
 	arith_circuit::ArithCircuitPoly,
-	memory::{PackedMemory, PackedMemorySliceMut},
+	memory::{PackedMemory, PackedMemorySlice, PackedMemorySliceMut},
 };
 
 /// Optimized CPU implementation of the compute layer.
@@ -176,14 +176,11 @@ impl<T: TowerFamily, P: PackedTop<T>> ComputeLayer<T::B128> for FastCpuLayer<T, 
 		slice: &mut <Self::DevMem as ComputeMemory<T::B128>>::FSliceMut<'_>,
 		value: T::B128,
 	) -> Result<(), Error> {
-		match slice {
-			PackedMemorySliceMut::Slice(items) => {
-				items.fill(P::broadcast(value));
-			}
-			PackedMemorySliceMut::SingleElement { owned, .. } => {
-				owned.fill(value);
-			}
-		};
+		let value = P::broadcast(value);
+
+		for element in slice.as_slice_mut() {
+			*element = value;
+		}
 		Ok(())
 	}
 }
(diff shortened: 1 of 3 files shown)


crates/fast_compute/src/layer.rs
```diff
@@ -176,14 +176,11 @@ impl<T: TowerFamily, P: PackedTop<T>> ComputeLayer<T::B128> for FastCpuLayer<T,
...
-			}
-			PackedMemorySliceMut::SingleElement { owned, .. } => {
-				owned.fill(value);
-			}
-		};
+		let value = P::broadcast(value);
+
+		for element in slice.as_slice_mut() {
+			*element = value;
+		}
```
GraDKh, line 183: Nice!
Can be simplified even further:
```
slice.as_slice_mut().fill(P::broadcast(value));
```

crates/fast_compute/src/layer.rs
```diff
@@ -622,11 +619,68 @@ impl<'a, T: TowerFamily, P: PackedTop<T>> ComputeLayerExecutor<T::B128>
...
+					input
+						.par_chunks(2)
+						.zip(round_output_data.as_slice_mut().par_iter_mut())
+						.for_each(|(chunk, output)| {
+							let scalar_iter = P::iter_slice(chunk)
+								.tuples()
+								.map(|(left, right)| left * right);
+							*output = P::from_scalars(scalar_iter);
+						});
+				}
```
GraDKh, line 671: Potentially it would be faster to re-use the packed multiplication by interleaving values:
```
let (lhs, rhs) = PackedField::interleave(chunk[0], chunk[1]);
let mults = lhs*rhs;

PackedField::from_scalars(mults.iter().step_by(2).copied(), mults.iter().skip(1).step_by(2).copied())
```

GraDKh (approved):

crates/fast_compute/src/layer.rs
```diff
@@ -622,11 +619,68 @@ impl<'a, T: TowerFamily, P: PackedTop<T>> ComputeLayerExecutor<T::B128>
...
+			Some(0) => {
+				return Err(Error::InputValidation(format!(
+					"input length must be greater than or equal to 2 in order to perform at least one reduction: {}",
+					input.len()
+				)));
+			}
+			Some(log_num_inputs) => log_num_inputs,
+		};
+		let expected_round_outputs_len = log_num_inputs;
+		if round_outputs.len() != expected_round_outputs_len as usize {
```
BenjaminTrapani, line 641: nit: maybe move the logic verifying the input and output dimensions to a separate helper function to share between the reference and fast implementations?

BenjaminTrapani (approved):

